PR #27, opened 2025-10-08
devx(release): Increase minor for update commits

NautiluX:
* Allow  as type in conventional commits
* Increase minor version on updates


siherrmann (approved):
Noice

PR #300, opened 2026-04-09
fix install codesphere

OliverTrautvetter:
[Clickup](https://app.clickup.com/t/869cukxue)


internal/bootstrap/gcp/install_config.go
```diff
@@ -247,16 +251,29 @@ func (b *GCPBootstrapper) UpdateInstallConfig() error {
...
-		b.Env.InstallConfig.Postgres.Primary.PrivateKey, b.Env.InstallConfig.Postgres.Primary.SSLConfig.ServerCertPem, err = installer.GenerateServerCertificate(
-			b.Env.InstallConfig.Postgres.CaCertPrivateKey,
-			b.Env.InstallConfig.Postgres.CACertPem,
-			b.Env.InstallConfig.Postgres.Primary.Hostname,
-			[]string{b.Env.InstallConfig.Postgres.Primary.IP})
-		if err != nil {
-			return fmt.Errorf("failed to generate primary server certificate: %w", err)
+		// Only regenerate postgres certificates if the IP or hostname changed,
+		// or if the private key was not loaded from the vault.
+		primaryNeedsRegen := b.Env.InstallConfig.Postgres.Primary.PrivateKey == "" ||
```
joka134: Doesn't need to be done in this PR necessarily, but maybe as part of this task: Can we move out the cert generation logic into a new funciton. the UpdateInstallConfig function is really long already 😄

internal/bootstrap/gcp/install_config_test.go
```diff
@@ -473,6 +473,131 @@ var _ = Describe("Installconfig & Secrets", func() {
...
+					csEnv.InstallConfig.Postgres.Primary.Hostname = "postgres"
+					csEnv.InstallConfig.Postgres.Primary.PrivateKey = key
+					csEnv.InstallConfig.Postgres.Primary.SSLConfig.ServerCertPem = cert
+				})
+
+				It("preserves existing cert/key without regeneration", func() {
+					origKey := csEnv.InstallConfig.Postgres.Primary.PrivateKey
+					origCert := csEnv.InstallConfig.Postgres.Primary.SSLConfig.ServerCertPem
+
+					// GenerateSecrets should NOT be called
```
joka134: What about something like `icg.EXPECT().GenerateSecrets().Return(nil).Times(0)`?

internal/bootstrap/gcp/install_config_test.go
```diff
@@ -473,6 +473,131 @@ var _ = Describe("Installconfig & Secrets", func() {
...
+
+					icg.EXPECT().WriteInstallConfig("fake-config-file", true).Return(nil)
+					icg.EXPECT().WriteVault("fake-secret", true).Return(nil)
+					nodeClient.EXPECT().CopyFile(mock.Anything, mock.Anything, mock.Anything).Return(nil).Twice()
+
+					err := bs.UpdateInstallConfig()
+					Expect(err).NotTo(HaveOccurred())
+
+					// IP should be updated to the node's IP
+					Expect(bs.Env.InstallConfig.Postgres.Primary.IP).To(Equal("10.0.0.1"))
```
joka134, line 542: Why do we expect `"10.0.0.1"`? Not really clear from the test case

joka134 (commented):
Nice! Left a few comments, but still approved

internal/bootstrap/gcp/install_config.go
OliverTrautvetter: done

internal/bootstrap/gcp/install_config_test.go
OliverTrautvetter: done

internal/bootstrap/gcp/install_config_test.go
OliverTrautvetter: I updated the comment

joka134 (approved):

PR #301, opened 2026-04-09
Positional arguments

siherrmann:
ClickUp: https://app.clickup.com/t/869bbxdg5


cli/cmd/root_test.go
```diff
@@ -0,0 +1,65 @@
...
+	return current
+}
+
+var _ = Describe("RootCmd", func() {
+	It("rejects positional args for commands configured with no positional args", func() {
+		rootCmd := cmd.GetRootCmd()
+		rootCmd.SilenceErrors = true
+		rootCmd.SilenceUsage = true
+
+		licensesCmd := findCommand(rootCmd, "licenses")
```
joka134: Can we build an independent test for `AddCmd`? With this, a change in licenses requires to find another example for this test case to work 🤔 Also it's not really clear what's tested here just from looking at this test code.

joka134 (changes requested):
Looks good, but one comment regarding the tests

cli/cmd/download_package.go
```diff
@@ -61,6 +61,7 @@ func AddDownloadPackageCmd(download *cobra.Command, opts *GlobalOptions) {
 			Short: "Download a codesphere package",
 			Long: io.Long(`Download a specific version of a Codesphere package
 				To list available packages, run oms list packages.`),
+			Args: cobra.ArbitraryArgs,
```
joka134, line 64: This doesn't affect how `PreRunE` is executed, right?

joka134: What about max args here as in your test case?

joka134 (approved):
Left two questions, but still approved 💯

PR #308, opened 2026-04-10
fix(bootstrap-local): Read service CIDR from local process

NautiluX:


internal/bootstrap/local/local.go
```diff
@@ -365,10 +397,37 @@ func (b *LocalBootstrapper) ReadClusterCIDRs() (podCIDR string, serviceCIDR stri
...
 	if serviceCIDR == "" {
-		return "", "", fmt.Errorf("could not determine service CIDR from kube-apiserver pod")
+		return "", fmt.Errorf("could not determine service CIDR from kube-apiserver pod")
 	}
 
-	return podCIDR, serviceCIDR, nil
+	return serviceCIDR, nil
+}
+
+// readServiceCIDRFromProd reads the service CIDR from the api server process on the local machine
```
schrodit:
```suggestion
// readServiceCIDRFromProc reads the service CIDR from the api server process on the local machine
```

internal/bootstrap/local/local.go
```diff
@@ -332,24 +333,55 @@ func (b *LocalBootstrapper) SyncCephMonEndpoints() error {
 // Pod CIDR is read from the first node's spec.podCIDR.
 // Service CIDR is read from the kube-apiserver pod's --service-cluster-ip-range flag.
```
schrodit: that comment is not correct anymore. we should update it that we also try to read it from the local process

internal/bootstrap/local/local.go
```diff
@@ -365,10 +397,37 @@ func (b *LocalBootstrapper) ReadClusterCIDRs() (podCIDR string, serviceCIDR stri
...
+		return "", fmt.Errorf("could not determine service CIDR from kube-apiserver pod")
 	}
 
-	return podCIDR, serviceCIDR, nil
+	return serviceCIDR, nil
+}
+
+// readServiceCIDRFromProd reads the service CIDR from the api server process on the local machine
+// this is necessary for single node k0s installations
+func (b *LocalBootstrapper) readServiceCIDRFromProc() (serviceCIDR string, err error) {
```
schrodit, line 418: This requires that the installer is called from the same machine where also the k0s installation runs.
Might make sense to include that in the error message.

internal/bootstrap/local/local.go
```diff
@@ -332,24 +333,55 @@ func (b *LocalBootstrapper) SyncCephMonEndpoints() error {
...
+	if serviceCIDR != "" {
+		return podCIDR, serviceCIDR, nil
+	}
+
+	log.Printf("can't read service CIDR from cluster, trying proc filesystem next: %s", err)
+
+	serviceCIDR, err = b.readServiceCIDRFromProc()
+
+	if err != nil {
+		err = fmt.Errorf("failed to determine service and/or pod CIDR. Pod CIDR: %s, service CIDR: %s, err: %s", podCIDR, serviceCIDR, err)
```
schrodit: Fine for now if we only support those 2 options.
To be distro independent it might make sense to be able to specify the service cidr as flag. but something for a follow up PR when we need it.

internal/bootstrap/local/local.go
NautiluX: added options to specify and put that as suggestion into the error message.

internal/bootstrap/local/local.go
NautiluX: updated

schrodit (approved):

PR #329, opened 2026-04-16
feat: add command to create a test user for automated testing

OliverTrautvetter:
[Clickup](https://app.clickup.com/t/869cuuda1)


cli/cmd/bootstrap_gcp.go
```diff
@@ -206,3 +216,38 @@ func writeInfraDetails(csEnv *gcp.CodesphereEnvironment) error {
...
+	}
+
+	pgHost := bs.Env.PostgreSQLNode.GetExternalIP()
+	if pgHost == "" {
+		return fmt.Errorf("postgres node has no external IP")
+	}
+
+	pgPassword := ""
+	if bs.Env.InstallConfig != nil {
+		pgPassword = bs.Env.InstallConfig.Postgres.AdminPassword
```
joka134: `bs.Env.InstallConfig.Postgres` can still be nil

internal/testuser/testuser.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	if opts.SSLMode == "" {
+		opts.SSLMode = DefaultSSLMode
+	}
+	if opts.Host == "" {
+		return nil, fmt.Errorf("host is required")
+	}
+	if opts.Password == "" {
+		return nil, fmt.Errorf("password is required")
+	}
```
joka134, line 75: I think a New(opts) function would be cleaner in this case, which stores the values and pings the DB. Create test user then starts with generateAPIToken()

internal/testuser/testuser.go
```diff
@@ -0,0 +1,261 @@
...
+
+	plaintextToken, err := generateAPIToken()
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate API token: %w", err)
+	}
+
+	hashedPassword := HashPassword(TestPassword)
```
joka134: Creating a random password would be better. Even if anyoing, the API's are public in most cases, so we should create strong password.

internal/testuser/testuser.go
```diff
@@ -0,0 +1,261 @@
...
+	defer func() { _ = db.Close() }()
+
+	db.SetConnMaxLifetime(30 * time.Second)
+	db.SetMaxOpenConns(1)
+
+	if err := db.Ping(); err != nil {
+		return nil, fmt.Errorf("failed to connect to database at %s:%d: %w", opts.Host, opts.Port, err)
+	}
+
+	log.Printf("Connected to PostgreSQL at %s:%d", opts.Host, opts.Port)
```
joka134, line 95: This should be in another function `connect`, that can be called after `New`

internal/testuser/testuser.go
```diff
@@ -0,0 +1,261 @@
...
+	err = tx.QueryRow(`
+		INSERT INTO authservice.credentials
+			(user_id, email, password_hash, authentication_method, signed_up, banned)
+		VALUES(nextval('authservice.credentials_user_id_seq'::regclass), $1, $2, 'password'::text, false, false)
+		RETURNING user_id`,
+		TestEmail, hashedPassword,
+	).Scan(&userID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to insert credentials: %w", err)
+	}
```
joka134, line 213: Also here: I think you split functionalities into small functions. One function per QueryRow, so that `createTestUserInDB` is creating the tx and calling the functions. It's easier to read and maintain.

internal/testuser/hash.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+)
+
+// Salts used by Codesphere's auth-service for password hashing.
+// Source: https://github.com/codesphere-cloud/codesphere-monorepo/blob/master/packages/auth-service/node/src/ts/config.ts#L14
+const (
+	salt1 string = "y6E5ZMkjRVtS"
```
joka134: Don't check these into git.

joka134 (changes requested):
First comments, still debugging the command a little more tomorrow

cli/cmd/bootstrap_gcp.go
```diff
@@ -181,3 +191,38 @@ func (c *BootstrapGcpCmd) BootstrapGcp() error {
...
+
+	if bs.Env.InstallConfig == nil {
+		return fmt.Errorf("install config not found in bootstrap environment")
+	}
+	pgPassword := bs.Env.InstallConfig.Postgres.AdminPassword
+	if pgPassword == "" {
+		return fmt.Errorf("postgres admin password not found in install config")
+	}
+
+	result, err := testuser.CreateTestUser(testuser.CreateTestUserOpts{
```
joka134:
```suggestion
	result, err := testuser.New(testuser.CreateTestUserOpts{
```
You should call the constructor from here. `testuser.CreateTestUser` is jsut another wrapper around the constructor. I think you don't need that. Just put the connect ping in New as well

internal/testuser/testuser.go
```diff
@@ -0,0 +1,344 @@
...
+		return nil, fmt.Errorf("test user %s already exists", TestEmail)
+	}
+
+	tx, err := c.db.Begin()
+	if err != nil {
+		return nil, fmt.Errorf("failed to begin transaction: %w", err)
+	}
+	committed := false
+	defer func() {
+		if !committed {
```
joka134: Why do you need a rollback if it's not commited?

internal/testuser/testuser.go
```diff
@@ -0,0 +1,344 @@
...
+// createInDB executes the database inserts inside a transaction.
+func (c *TestUserCreator) createInDB(hashedPassword, hashedToken string) (*TestUserResult, error) {
+	var exists bool
+	err := c.db.QueryRow(`SELECT EXISTS(SELECT 1 FROM authservice.credentials WHERE email = $1)`, TestEmail).Scan(&exists)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check for existing test user: %w", err)
+	}
+	if exists {
+		return nil, fmt.Errorf("test user %s already exists", TestEmail)
+	}
```
joka134: new function? 😄

internal/testuser/testuser.go
```diff
@@ -0,0 +1,344 @@
...
+		TestEmail, hashedPassword,
+	).Scan(&userID)
+	if err != nil {
+		return 0, fmt.Errorf("failed to insert credentials: %w", err)
+	}
+	return userID, nil
+}
+
+func (c *TestUserCreator) insertEmailConfirmation(tx *sql.Tx) error {
+	emailConfirmationIDBytes := make([]byte, 16)
```
joka134: What about [uuid](https://github.com/google/uuid) or letting postgres do it? Seems much easier to me:

`INSERT INTO authservice.email_confirmations
(id, email, pending, created_at)
VALUES(uuid_generate_v4(), 'test@codesphere.com', false, CURRENT_TIMESTAMP);`

cli/cmd/bootstrap_gcp.go
```diff
@@ -163,7 +166,14 @@ func (c *BootstrapGcpCmd) BootstrapGcp() error {
...
 
 	log.Println("\n🎉🎉🎉 GCP infrastructure bootstrapped successfully!")
-	log.Printf("Access the jumpbox using:\nssh-add $SSH_KEY_PATH; ssh -o StrictHostKeyChecking=no -o ForwardAgent=yes -o SendEnv=OMS_PORTAL_API_KEY root@%s", bs.Env.Jumpbox.GetExternalIP())
+	log.Printf("Access the jumpbox using:\nssh-add $SSH_KEY_PATH; ssh -o StrictHostKeyChecking=no -o ForwardAgent=yes -o SendEnv=OMS_PORTAL_API_KEY -o SendEnv=OMS_PORTAL_API root@%s", bs.Env.Jumpbox.GetExternalIP())
+
+	if c.CreateTestUser {
+		if err := c.createTestUser(bs); err != nil {
+			log.Printf("warning: failed to create test user: %v", err)
+		}
+	}
```
joka134: Would it make sense to call this inside the bootstrap package at the very end?

internal/testuser/hash.go
```diff
@@ -0,0 +1,39 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"os"
+)
+
+// HashPassword hashes a password using Codesphere's double-SHA256 scheme with salts.
+// The salts are read from environment variables SALT_1 and SALT_2.
+func HashPassword(password string) (string, error) {
+	salt1 := os.Getenv("SALT_1")
```
joka134:
```suggestion
	salt1 := os.Getenv("OMS_CS_SALT_1")
```
env vars can be a bit more specific for what they are used

joka134: alternative is flags, but with this we at least don't need to show the salt on every cli execution 🤷

joka134 (changes requested):
Nice, much cleaner. Still some comments going into a similar direction

internal/installer/node/node.go
```diff
@@ -89,11 +90,15 @@ func (r *SSHNodeClient) RunCommand(n *Node, username string, command string) err
 	defer util.IgnoreError(session.Close)
 
 	_ = session.Setenv("OMS_PORTAL_API_KEY", os.Getenv("OMS_PORTAL_API_KEY"))
+	_ = session.Setenv("OMS_PORTAL_API", os.Getenv("OMS_PORTAL_API"))
 	_ = agent.RequestAgentForwarding(session) // Best effort, ignore errors
 
+	var stderrBuf bytes.Buffer
```
joka134, line 96:
```suggestion
	var stderrBuf bytes.Buffer
	session.Stderr = &stderrBuf
```

and drop the else statement

internal/installer/node/node.go
```diff
@@ -89,11 +90,15 @@ func (r *SSHNodeClient) RunCommand(n *Node, username string, command string) err
 	defer util.IgnoreError(session.Close)
 
 	_ = session.Setenv("OMS_PORTAL_API_KEY", os.Getenv("OMS_PORTAL_API_KEY"))
+	_ = session.Setenv("OMS_PORTAL_API", os.Getenv("OMS_PORTAL_API"))
```
joka134, line 93: Why is needed for test user creation?

internal/testuser/testuser.go
```diff
@@ -0,0 +1,323 @@
...
+		_ = db.Close()
+		return nil, fmt.Errorf("failed to connect to database at %s:%d: %w", opts.Host, opts.Port, err)
+	}
+
+	log.Printf("Connected to PostgreSQL at %s:%d", opts.Host, opts.Port)
+	return &TestUserCreator{opts: opts, db: db}, nil
+}
+
+// newWithDB creates a TestUserCreator with an already-opened DB.
+func newWithDB(opts CreateTestUserOpts, db *sql.DB) *TestUserCreator {
```
joka134: Do we need this? In the tests you can also just use &TestUserCreator{opts: opts, db: db}, right

internal/testuser/hash.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"os"
+)
+
+// HashPassword hashes a password using Codesphere's double-SHA256 scheme with salts.
+// The salts are read from environment variables OMS_CS_SALT_1 and OMS_CS_SALT_2.
+func HashPassword(password string) (string, error) {
```
joka134: Maybe I came to a new conclusion to just keep out the salts of oms and create a password hash manually, that is provided via env. That means the test user always gets the same password, which is fine as signup is usually open for test instances. And now one will be forced to copy/paste salts to local environments 😆 
The password (+ hash for local development) will be stored in 1password

joka134 (changes requested):
some small things + the one thing we discussed

internal/installer/node/node.go
OliverTrautvetter: One time I got some weird error that OMS_PORTAL_API was not set. I couldn't reproduce the problem, but I added this line to node. Not sure if this problem happened randomly though

internal/testuser/hash.go
```diff
@@ -0,0 +1,20 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+)
+
+// HashAPIToken hashes an API token using a single SHA256 with no additional salt.
+func HashAPIToken(apiToken string) string {
+	return hashSecret(apiToken, "")
+}
+
+func hashSecret(secret, salt string) string {
```
joka134:
```suggestion
func hashSecret(secret string) string {
```
We don't need it anymore, right?

joka134 (changes requested):

internal/testuser/hash.go
OliverTrautvetter: yes, removed

joka134 (approved):
🥇

PR #330, opened 2026-04-16
feat: write infra file after gcp project creation

joka134:
- Move infra file functions into gcp package
- Write infra file for the first time just after project creation

This PR saves some GCP infra details in the `gcp-infra.json` after creating the GCP project, so that `bootstrap-gcp cleanup` can be used in case of an unexpected termination.

https://app.clickup.com/t/869cn7e4k


internal/bootstrap/gcp/infrafile.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal codesphere env: %w", err)
+	}
+
+	workdir := env.NewEnv().GetOmsWorkdir()
+	fw := util.NewFilesystemWriter()
+
+	err = fw.MkdirAll(workdir, 0755)
+	if err != nil {
+		return fmt.Errorf("failed to create workdir %w", err)
```
OliverTrautvetter:
```suggestion
		return fmt.Errorf("failed to create workdir: %w", err)
```

internal/bootstrap/gcp/infrafile.go
```diff
@@ -0,0 +1,63 @@
...
+
+// WriteInfraFile writes details about the bootstrapped codesphere environment into a file.
+func (b *GCPBootstrapper) WriteInfraFile() error {
+	envBytes, err := json.MarshalIndent(b.Env, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal codesphere env: %w", err)
+	}
+
+	workdir := env.NewEnv().GetOmsWorkdir()
+	fw := util.NewFilesystemWriter()
```
OliverTrautvetter: we should not create a new fw, we should already have b.fw

internal/bootstrap/gcp/gcp.go
```diff
@@ -216,6 +190,11 @@ func (b *GCPBootstrapper) Bootstrap() error {
 		return fmt.Errorf("failed to ensure GCP project: %w", err)
 	}
 
+	err = b.stlog.Step("Write infrastructure file", b.WriteInfraFile)
+	if err != nil {
+		return fmt.Errorf("failed to write infrastructure file: %w", err)
+	}
```
OliverTrautvetter, line 196: This error aborts the entire bootstrap process, even though the GCP project has already been created at this point. Do we really want this? Before it was just a warning

cli/cmd/bootstrap_gcp.go
```diff
@@ -149,18 +148,14 @@ func (c *BootstrapGcpCmd) BootstrapGcp() error {
 		c.CodesphereEnv.FeatureFlags[flag] = true
 	}
 
-	bootstrapErr := bs.Bootstrap()
+	err = bs.Bootstrap()
+	writeGcpInfraFile(bs)
```
OliverTrautvetter: In Bootstrap() we do 
	err = b.stlog.Step("Write infrastructure file", b.WriteInfraFile)

in writeGcpInfraFile(bs)
if err != nil {
		log.Printf("warning: failed to write infra details: %v", err)
	} else {
		log.Printf("infrastructure details written to file: %s", gcp.GetInfraFilePath())
	}

I think we can remove the log in Bootstrap() and only have the final log in writeGcpInfraFile

OliverTrautvetter: Also both have a different error handling. 
writeGcpInfraFile -> log
Bootstrap -> Errorf

internal/bootstrap/gcp/infrafile.go
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Codesphere Inc.
```
OliverTrautvetter, line 1: no unit tests for this file?

OliverTrautvetter (commented):
nice 👍 some comments

internal/bootstrap/gcp/infrafile.go
joka134: I was just splitting the functions into a new file and moving the write function into this package. We didn't have tests so far. But good point, I will add some tests

joka134: Done

cli/cmd/bootstrap_gcp.go
joka134: Removing the log in bootstrap would not make it transparent to the user that this has happend. In general I think we should log both.

internal/bootstrap/gcp/gcp.go
joka134: Yeah, thought about it, but I think it's the most consistent way. Th potential to fail is not really high, only if you don't have permissions to write files. And in that case it would also fail later when creating the config.yaml 🤷

OliverTrautvetter: That's probably true, it would just fail later on. It would be strange from the start not to have permission to write files 😄

cli/cmd/bootstrap_gcp.go
OliverTrautvetter: Fine for me, this was king of nit

internal/bootstrap/gcp/gcp.go
joka134: But we should act in the same way in both cases. So fail or log error in both cases. Logging an error is not really visible with the step logger actually

cli/cmd/bootstrap_gcp.go
```diff
@@ -149,26 +148,28 @@ func (c *BootstrapGcpCmd) BootstrapGcp() error {
...
 	}
 
-	bootstrapErr := bs.Bootstrap()
+	err = bs.Bootstrap()
 
-	err = writeInfraDetails(bs.Env)
-	if err != nil {
-		log.Printf("warning: failed to write infra details: %v", err)
+	writeInfraErr := bs.WriteInfraFile()
+	if writeInfraErr != nil {
```
OliverTrautvetter, line 154: 👀 we should check err before we do some other stuff

OliverTrautvetter (approved):
Looks good now, just one comment, still approved 🚀

cli/cmd/bootstrap_gcp.go
joka134: This is the behaviour just like before. We try to write the infra file before actually checking if bootstrapping was successfull

PR #331, opened 2026-04-16
chore: remove duplicated tests

joka134:
Tests have been moved to `install_config.go`, but are not removed in the old file.


OliverTrautvetter (approved):
Good catch, we do that already in install_config_test.go 😄

PR #353, opened 2026-04-28
feat: add random email generation for test users in TestUserCreator

OliverTrautvetter:
To prevent "warning: failed to create test user: test user test@codesphere.com already exists" during the installation in PC-Testing GCP


joka134:
Is it causing some issues to get this warning? I'm fine with generating random mail users, just asking because I don't see the need at the moment. it will also create a new user on every re-run/update, right?

joka134 (approved):

PR #370, opened 2026-05-08
fix: remove pgoperator clusterwide overwrite

joka134:
https://codesphere-cloud.slack.com/archives/C03DDNG83D2/p1778243099976459?thread_ts=1778242537.292149&cid=C03DDNG83D2

### DIFF
--- a/internal/installer/resource_profiles.go
+++ b/internal/installer/resource_profiles.go
@@ -175,9 +175,6 @@ func applyNoRequestsProfile(config *files.RootConfig) {
 	}
 	config.Cluster.PgOperator.Override = util.DeepMergeMaps(config.Cluster.PgOperator.Override, map[string]any{
 		"cloudnative-pg": map[string]any{
-			"config": map[string]any{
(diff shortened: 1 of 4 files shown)


internal/installer/resource_profiles.go
```diff
@@ -175,9 +175,6 @@ func applyNoRequestsProfile(config *files.RootConfig) {
 	}
 	config.Cluster.PgOperator.Override = util.DeepMergeMaps(config.Cluster.PgOperator.Override, map[string]any{
 		"cloudnative-pg": map[string]any{
-			"config": map[string]any{
```
utkuerol, line 178: why was it explicitly overriden to false in the first place? 🤔  i don't see how it would be seen relevant for resource profiling

utkuerol (approved):

NautiluX (approved):

internal/installer/resource_profiles.go
joka134: I don't know 🤷

PR #372, opened 2026-05-11
feat: add LTS 1.77.2 compatibility and enhance related functionalities

OliverTrautvetter:
**LTS Installer Compatibility and OMS Binary Handling:**

* Added logic to detect LTS installer versions in `InstallCodesphere`, ensuring that for such versions, a freshly-built OMS binary is copied to the jumpbox and a Ceph master host watcher process is started/stopped appropriately. This ensures compatibility with LTS-specific requirements.
* Introduced the `OmsBinaryBuilder` field to `GCPBootstrapper` (with default assignment to `BuildOmsLinuxBinary`) to allow injection of custom OMS binary build logic, improving testability and flexibility.
* Extended tests to cover LTS installer scenarios, including OMS binary update and error handling paths.

**Install Configuration and SSH Port Consistency:**

* Ensured all generated install config entries for Postgres, Ceph, and Kubernetes nodes explicitly set `SSHPort` to `22`, improving clarity and compatibility with SSH routines.
* Updated the install config to record the `GeneratedForVersion` field, tracking the installer version the config was generated for.
* For older LTS installers, added logic to write a compatibility-stripped config file for the jumpbox when required, and to use this file when copying to the jumpbox.

[Clickup](https://app.clickup.com/t/869d13gx4)


cli/cmd/download_package.go
```diff
@@ -101,37 +101,47 @@ func (c *DownloadPackageCmd) DownloadBuild(p portal.Portal, build portal.Build,
...
+		if fileInfo, statErr := out.Stat(); statErr == nil {
+			fileSize = int(fileInfo.Size())
+		}
 
-	err = p.DownloadBuildArtifact("codesphere", download, out, fileSize, c.Opts.Quiet)
-	if err != nil {
-		return fmt.Errorf("failed to download build: %w", err)
-	}
+		err = p.DownloadBuildArtifact("codesphere", download, out, fileSize, c.Opts.Quiet)
+		util.CloseFileIgnoreError(out)
```
joka134: defer after creating the file is the safest way

cli/cmd/download_package.go
```diff
@@ -101,37 +101,47 @@ func (c *DownloadPackageCmd) DownloadBuild(p portal.Portal, build portal.Build,
...
-	if err != nil {
-		return err
-	}
-	defer util.CloseFileIgnoreError(verifyFile)
+		verifyFile, err := c.FileWriter.Open(fullFilename)
+		if err != nil {
+			return err
+		}
+		verifyErr := p.VerifyBuildArtifactDownload(verifyFile, download)
+		util.CloseFileIgnoreError(verifyFile)
```
joka134: same here

cli/cmd/download_package.go
joka134, line 1: Seems unrelated to the LTS changes. Can we move them to smaller PRs. I would approve this file for example, but want to debug bootstrapping a bit, which will take some time.

cli/cmd/download_package.go
```diff
@@ -101,37 +101,47 @@ func (c *DownloadPackageCmd) DownloadBuild(p portal.Portal, build portal.Build,
 	}
 
 	fullFilename := build.BuildPackageFilename(filename)
-	out, err := c.FileWriter.OpenAppend(fullFilename)
-	if err != nil {
-		out, err = c.FileWriter.Create(fullFilename)
+	for retried := false; ; retried = true {
```
joka134: Is this retrying forever? You want to auto-solve invalid partial packages, right? What if network is down? How often is this actually happening?

cli/cmd/download_package.go
```diff
@@ -101,37 +101,47 @@ func (c *DownloadPackageCmd) DownloadBuild(p portal.Portal, build portal.Build,
...
+		if verifyErr == nil {
+			return nil
+		}
 
-	return nil
+		if !retried && fileSize > 0 {
+			// Resumed from a partial file but verification failed, stale or corrupt bytes.
+			// Delete the partial file and retry the full download from scratch.
+			log.Println("Verification failed on resumed download; retrying from scratch...")
+			if removeErr := c.FileWriter.Remove(fullFilename); removeErr == nil {
```
joka134: Do we want to delete the package automatically? This could also be a hint for the user to clean up manually 🤔

joka134 (changes requested):
Please split the PR. Just reviewed the download package part so far. Will continue of course, but wanted to leave the feedback on that feature already.

cli/cmd/download_package.go
OliverTrautvetter: Is there a reason not to delete it automatically? Or should we maybe log a warning and then ask if the user would like to delete it y/n ?

cli/cmd/download_package.go
OliverTrautvetter: Moved to a new PR https://github.com/codesphere-cloud/oms/pull/421

internal/bootstrap/gcp/gcp.go
```diff
@@ -954,7 +983,13 @@ func (b *GCPBootstrapper) ensureCodespherePackageOnJumpbox() (string, error) {
...
 	err := b.Env.Jumpbox.RunSSHCommand("root", downloadCmd)
 	if err != nil {
-		return "", fmt.Errorf("failed to download Codesphere package from jumpbox: %w", err)
+		// A stale partial file from a previous (different) build can cause MD5 verification to
+		// fail even after a successful byte-range resume. Delete it and retry from scratch.
+		b.stlog.Logf("Download failed; removing any stale partial file and retrying from scratch...")
+		cleanAndRetryCmd := fmt.Sprintf("rm -f %s && %s", fullPackageFilename, downloadCmd)
+		if retryErr := b.Env.Jumpbox.RunSSHCommand("root", cleanAndRetryCmd); retryErr != nil {
+			return "", fmt.Errorf("failed to download Codesphere package from jumpbox: %w", retryErr)
+		}
```
joka134: after the discussion today: this should simply fail

internal/bootstrap/gcp/lts.go
```diff
@@ -0,0 +1,122 @@
...
+
+// ltsRegistry is the single source of truth for all known LTS versions and their quirks.
+// Add a new LTSSpec entry here to support an additional LTS release.
+var ltsRegistry = []LTSSpec{
+	{
+		InstallVersion: "codesphere-lts-v1.77.2",
+		UnsupportedExperiments: []string{
+			"secret-management",
+			"sub-path-mount",
+		},
```
joka134: That means we need to maintain the list when adding new experiments? What about listing all supported experiments? Then we don't need to touch it anymore

internal/bootstrap/gcp/lts.go
```diff
@@ -0,0 +1,122 @@
...
+		ClearManagedServices:    true,
+		RequiresJumpboxFiles:    true,
+		RequiresOmsBinaryUpdate: true,
+	},
+}
+
+// FindLTSSpec returns the LTSSpec for the given installVersion, or nil if it is not a
+// known LTS release that requires special handling.
+func FindLTSSpec(installVersion string) *LTSSpec {
```
joka134, line 80: why exported?

internal/bootstrap/gcp/lts.go
```diff
@@ -0,0 +1,122 @@
...
+	// and codesphere.yaml (with compat applied) instead of using the standard config.yaml.
+	RequiresJumpboxFiles bool
+	// RequiresOmsBinaryUpdate instructs the bootstrap to build a fresh linux/amd64 OMS binary
+	// and deploy it to the jumpbox before running the installer.
+	RequiresOmsBinaryUpdate bool
+}
+
+// ltsRegistry is the single source of truth for all known LTS versions and their quirks.
+// Add a new LTSSpec entry here to support an additional LTS release.
+var ltsRegistry = []LTSSpec{
```
joka134, line 60: const?

cli/cmd/update_dockerfile_test.go
joka134, line 1: Did this cause some issues? Why changing it in this PR?

docs/oms_beta_bootstrap-gcp.md
```diff
@@ -29,7 +29,7 @@ oms beta bootstrap-gcp [flags]
       --datacenter-name string                  Datacenter name (default: dev) (default "dev")
       --dns-project-id string                   GCP Project ID for Cloud DNS (optional)
       --dns-zone-name string                    Cloud DNS Zone Name (optional) (default "oms-testing")
-      --experiments stringArray                 Experiments to enable in Codesphere installation (optional) (default [managed-services,headless-services,vcluster,custom-service-image,ms-in-ls,secret-management,sub-path-mount])
```
joka134: Why disable them by default? Aren't they required for new versions?

internal/bootstrap/gcp/gcp.go
```diff
@@ -917,6 +917,12 @@ func (b *GCPBootstrapper) InstallCodesphere() error {
 		return fmt.Errorf("failed to ensure Codesphere package on jumpbox: %w", err)
 	}
 
+	if ltsSpec := FindLTSSpec(b.Env.InstallVersion); ltsSpec != nil && ltsSpec.RequiresOmsBinaryUpdate {
+		if err := b.ensureNewOmsBinaryOnJumpbox(); err != nil {
+			return fmt.Errorf("failed to update OMS binary on jumpbox for %s: %w", b.Env.InstallVersion, err)
+		}
+	}
```
joka134: Why updating the oms binary for LTS? If only for development then we should do it for all versions. But let's move that feature to another PR

.gitignore
```diff
@@ -48,6 +48,8 @@ internal/util/testdata/
 config.yaml
 prod.vault.yaml
 configure-k0s.sh
+config-jumpbox.yaml
+codesphere.yaml
```
joka134: If these is the config.yaml for lts it will be confusing for the users, why they have 2 or 3 configs laying around. What we should do:
- Adding CS version info to the config.yaml
- Add suffix for LTS versions `config-lts-1_77_2.yaml`  or something like this

That would mean we generate a new config for LTS, but don't overwrite newer config.yaml that may still exist. I think that's what you wanted achive with this

internal/installer/config_manager_profile.go
```diff
@@ -177,10 +177,98 @@ func (g *InstallConfig) applyCommonProperties() {
...
+				IconURL:     "/ide/assets/managed-services/postgresql.svg",
+				Scope:       "global",
+				Capabilities: &files.ManagedServiceCapabilities{
+					Pause:               true,
+					Backups:             true,
+					PointInTimeRecovery: true,
+				},
+				// ConfigSchema: ,
+				// DetailsSchema
+				// secretsSchema
```
joka134: Are these TODOs?

internal/installer/config_manager_profile.go
```diff
@@ -177,10 +177,98 @@ func (g *InstallConfig) applyCommonProperties() {
...
+			{Name: "s3", Version: "v1", Backend: files.ManagedServiceBackend{
+				API: files.ManagedServiceAPI{
+					Endpoint: "http://localhost:8080",
+				},
+			}},
+			{Name: "virtual-k8s", Version: "v1", Backend: files.ManagedServiceBackend{
+				API: files.ManagedServiceAPI{
+					Endpoint: "http://localhost:8080",
+				},
+			}},
```
joka134: We should also add the default values from the monorepo here: https://github.com/codesphere-cloud/codesphere-monorepo/blob/master/packages/marketplace/node/src/ts/providers.ts

joka134 (changes requested):

internal/bootstrap/gcp/lts.go
OliverTrautvetter: go does not support declaring slices or arrays as constants, but ltsRegistry is not exported, so it should be fine

internal/bootstrap/gcp/lts.go
OliverTrautvetter: for using it in lts_test.go, gcp.go and install_config.go

internal/bootstrap/gcp/gcp.go
```diff
@@ -964,6 +966,16 @@ func (b *GCPBootstrapper) InstallCodesphere() error {
 		return fmt.Errorf("failed to ensure Codesphere package on jumpbox: %w", err)
 	}
 
+	if ltsSpec := FindLTSSpec(b.Env.InstallVersion); ltsSpec != nil {
+		if ltsSpec.RequiresOmsBinaryUpdate {
+			if err := b.ensureNewOmsBinaryOnJumpbox(); err != nil {
+				return fmt.Errorf("failed to update OMS binary on jumpbox for %s: %w", b.Env.InstallVersion, err)
+			}
+		}
+		b.startLTSCephMasterWatcher()
```
joka134: Which LTS versions need this? Is this fixed on current master?

internal/bootstrap/gcp/gcp.go
```diff
@@ -972,6 +984,61 @@ func (b *GCPBootstrapper) InstallCodesphere() error {
...
+
+	return nil
+}
+
+// startLTSCephMasterWatcher starts a background process on the ceph master node that continuously
+// re-adds the master to the Ceph orchestrator host inventory. This is required for LTS versions
+// because the installer's configureHosts step applies a declarative host spec containing only the
+// non-master nodes, which removes the master from the inventory. The watcher restores it within
+// seconds, before the subsequent configureMonitors step runs.
+func (b *GCPBootstrapper) startLTSCephMasterWatcher() {
```
joka134, line 1583: So we are fighting a bug in the 1.77.2 LTS release here? 🙈

internal/bootstrap/gcp/lts.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+// LTSSpec describes the compatibility requirements for a specific LTS release.
+// To add support for a new LTS version, add a new entry to ltsRegistry — no other
+// files need to change.
+type LTSSpec struct {
+	// InstallVersion is the exact install version string that identifies this LTS release.
+	InstallVersion string
+	// Experiments lists all experiments supported by this LTS release.
+	// Only these experiments are passed to the installer, any others are stripped.
```
joka134: So we are ignoring the one set in the experiments flag? I think we should throw an error if there is any unsupported experiment set by flags.

internal/bootstrap/gcp/lts.go
```diff
@@ -0,0 +1,122 @@
...
+	Experiments []string
+	// ClearManagedServices instructs the compat layer to set ManagedServices to nil.
+	// Required when the LTS schema expects full provider definitions, not the abbreviated
+	// form stored in config.yaml.
+	ClearManagedServices bool
+	// RequiresJumpboxFiles instructs the bootstrap to generate LTS-versioned compat config files
+	// (e.g. config-lts-1_77_2.yaml and codesphere-lts-1_77_2.yaml) instead of using config.yaml directly.
+	RequiresJumpboxFiles bool
```
joka134, line 32: Isn't this true for all LTS versions? Why make it variable if we need it for all?

internal/bootstrap/gcp/lts.go
```diff
@@ -0,0 +1,122 @@
...
+	// RequiresOmsBinaryUpdate instructs the bootstrap to build a fresh linux/amd64 OMS binary
+	// and deploy it to the jumpbox before running the installer.
+	RequiresOmsBinaryUpdate bool
```
joka134, line 37: Why do we need this?

internal/bootstrap/gcp/lts.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	return jumpboxConfigBytes, nil
+}
+
+// ApplyLTSCompat adjusts a CodesphereConfig to be compatible with the given LTS release:
+//  1. Only experiments listed in the LTS spec are kept; all others are stripped.
+//  2. ManagedServices is cleared when the LTS schema requires full provider definitions.
+func ApplyLTSCompat(cfg *files.CodesphereConfig, spec *LTSSpec) {
+	cfg.Experiments = FilterExperiments(cfg.Experiments, spec.Experiments)
```
joka134, line 137: Can we throw an error if some experiments have been set in the flags that are not supported by the version?

internal/bootstrap/gcp/lts.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil, fmt.Errorf("failed to marshal %s jumpbox config: %w", spec.InstallVersion, err)
+	}
+
+	return jumpboxConfigBytes, nil
+}
+
+// ApplyLTSCompat adjusts a CodesphereConfig to be compatible with the given LTS release:
+//  1. Only experiments listed in the LTS spec are kept; all others are stripped.
+//  2. ManagedServices is cleared when the LTS schema requires full provider definitions.
+func ApplyLTSCompat(cfg *files.CodesphereConfig, spec *LTSSpec) {
```
joka134: All of these functions can be unexported

internal/bootstrap/gcp/gcp.go
joka134: If yes, we can have a flag for the lts version if this needs to be executed

joka134 (changes requested):
Bootstrapping works, but managed services are not configured. Also as discussed some comments.

PR #384, opened 2026-05-11
feat(bootstrap): add other git providers to oms bootstrap command

NJona:
Adding gitlab, bitbucket, azure dev ops to bootstrap command.

[Clickup](https://app.clickup.com/t/869d7h9wd)

Related PR:
- Adding git providers to pc deploy in monorepo: https://github.com/codesphere-cloud/codesphere-monorepo/pull/19280


internal/bootstrap/gcp/install_config.go
```diff
@@ -242,6 +242,61 @@ func (b *GCPBootstrapper) UpdateInstallConfig() error {
 			},
 		}
 	}
+	if b.Env.GitLabAppClientID != "" && b.Env.GitLabAppClientSecret != "" {
```
joka134, line 245: In `gcp.go` is a `validateGitHubParams` to check if either none or all values for GH are set. We should do a verify step for the new providers as well to ensure that if clientID is set and secret is not, we throw an error. Else it's simply ignored, which could be annoying when realizing it's not applied after waiting for the bootstrap to finish.

joka134 (changes requested):
Nice, just one comment to add some flag validation.

internal/bootstrap/gcp/install_config.go
NJona: added the validation 👍

joka134 (approved):
🔥

PR #401, opened 2026-05-13
feat(gcp): Make external loki configurable

schrodit:


internal/bootstrap/gcp/gcp.go
```diff
@@ -486,6 +494,18 @@ func (b *GCPBootstrapper) validateOidcParams() error {
 	return nil
 }
 
+func (b *GCPBootstrapper) validateExternalLokiParams() error {
+	if b.Env.ExternalLokiEndpoint != "" {
+		return nil
+	}
+
+	if b.Env.ExternalLokiSecret != "" || b.Env.ExternalLokiUser != "" {
+		return fmt.Errorf("external Loki endpoint is required when external Loki secret or user is set")
```
joka134: ExternalLokiEndpoint is set at this point. So error is more like the other way around
```suggestion
		return fmt.Errorf("external Loki secret and user is required when external loki endpoint is set")
```

joka134 (approved):
lgtm

PR #410, opened 2026-05-15
feat(bootstrap): add Central otel credentials options

BBesrour:
CU-869d79gqh

Adds possibility to pass central otel creds. Need to send traces to central collector (deployed on dc0) so that they are viewable in central grafana.

Related PRs:
- https://github.com/codesphere-cloud/charts/pull/554
- https://github.com/codesphere-cloud/codesphere-monorepo/pull/19336


NautiluX (approved):

internal/installer/files/config_yaml.go
```diff
@@ -340,6 +340,12 @@ type OpenBaoConfig struct {
 	Password string `yaml:"-"`
 }
 
+type CentralOtelCredentials struct {
```
schrodit: Mit: those are not only credentials but enable the whole component.
Maybe simple CentralOtelConfig is better suited

BBesrour: Also just noticed that we need to add the Override property there

PR #413, opened 2026-05-15
update(deps): bump cs-go to v1.1.1 (traceIds in CS API error logs)

nathanael-ruf:


go.mod
```diff
@@ -51,7 +51,6 @@ require (
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.2
 	github.com/google/go-github/v74 v74.0.0
-	github.com/google/go-github/v86 v86.0.0
```
nathanael-ruf, line 54: `go mod tidy` removed this. Not sure why was even here in the first place then.

PR #430, opened 2026-05-19
fix(openbao): improve install command

Jcing95:
## Summary

- Add `--namespace` / `-n` flag to `oms install openbao` to allow deploying OpenBao to a custom Kubernetes namespace (defaults to `vault` for backward compatibility)
- Fix multi-namespace support: the Bank-Vaults Operator is cluster-scoped, so skip re-deployment if it already exists in another namespace (prevents Helm ownership conflicts on ClusterRole resources)
- Ensure the target namespace is created before applying the Vault CR, since Helm no longer implicitly creates it when the operator lives elsewhere
- Fix unchecked `os.Setenv` return value flagged by the linter

### DIFF
--- a/cli/cmd/install_openbao.go
+++ b/cli/cmd/install_openbao.go
@@ -150,7 +150,7 @@ func AddInstallOpenBaoCmd(install *cobra.Command, opts *GlobalOptions) {
 	openbao.cmd.Flags().StringVar(&openbao.Opts.StorageSize, "storage-size", "10Gi", "PVC storage size for each OpenBao replica")
 	openbao.cmd.Flags().DurationVar(&openbao.Opts.Timeout, "timeout", 5*time.Minute, "Timeout for waiting on initialization")
 	openbao.cmd.Flags().StringVarP(&openbao.Opts.AgeKeyFile, "age-key-file", "k", "", "Path to age private key file for SOPS encryption/decryption (auto-detected if not set)")
-	openbao.cmd.Flags().BoolVarP(&openbao.Opts.Yes, "yes", "y", false, "Auto-approve fresh initialization when no DR backup is found")
+	openbao.cmd.Flags().BoolVarP(&openbao.Opts.Yes, "yes", "y", false, "Auto-approve re-initialization of an existing deployment when no DR backup is found")
 
 	util.MarkFlagRequired(openbao.cmd, "dr-backup-path")
 
--- a/cli/cmd/install_openbao.go
+++ b/cli/cmd/install_openbao.go
@@ -4,12 +4,16 @@
 package cmd
 
 import (
+	"bufio"
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
@@ -35,11 +39,19 @@ type InstallOpenBaoOpts struct {
 	Replicas          int
 	StorageSize       string
 	Timeout           time.Duration
+	AgeKeyFile        string
+	Yes               bool
 }
 
 func (c *InstallOpenBaoCmd) RunE(_ *cobra.Command, _ []string) error {
 	if err := validateOpenBaoPrereqs(); err != nil {
 		return err
+	}
+
+	// If --age-key-file is provided, set SOPS_AGE_KEY_FILE so ResolveAgeKey
+	// picks it up. Otherwise, fall back to the normal auto-discovery chain.
+	if c.Opts.AgeKeyFile != "" {
+		os.Setenv("SOPS_AGE_KEY_FILE", c.Opts.AgeKeyFile)
 	}
 
 	configDir, err := os.UserConfigDir()
@@ -67,6 +79,32 @@ func (c *InstallOpenBaoCmd) RunE(_ *cobra.Command, _ []string) error {
 	inst, err := installer.NewOpenBaoInstaller(cfg)
 	if err != nil {
 		return fmt.Errorf("initializing openbao installer: %w", err)
+	}
+
+	inst.ConfirmFunc = func() error {
+		if c.Opts.Yes {
+			return nil
+		}
+
+		fmt.Printf("\nWARNING: No DR backup found at: %s\n", c.Opts.DRBackupPath)
+		fmt.Println("This will perform a FRESH OpenBao initialization:")
+		fmt.Println("  - Existing Vault CR will be deleted")
+		fmt.Println("  - All OpenBao pods will be terminated")
+		fmt.Println("  - Persistent volume claims (data) will be deleted")
+		fmt.Println("  - Existing unseal keys will be removed")
+		fmt.Println("")
+		fmt.Println("If you intended to restore from a backup, verify --dr-backup-path is correct.")
+		fmt.Print("\nType 'yes' to continue: ")
+
+		reader := bufio.NewReader(os.Stdin)
+		input, err := reader.ReadString('\n')
+		if err != nil && !errors.Is(err, io.EOF) {
+			return fmt.Errorf("failed to read confirmation: %w", err)
+		}
+		if strings.TrimSpace(strings.ToLower(input)) != "yes" {
+			return fmt.Errorf("aborted: type 'yes' to continue or pass --yes")
+		}
+		return nil
 	}
 
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
@@ -106,6 +144,8 @@ func AddInstallOpenBaoCmd(install *cobra.Command, opts *GlobalOptions) {
 	openbao.cmd.Flags().IntVar(&openbao.Opts.Replicas, "replicas", 1, "Number of OpenBao replicas (1 for single-node, odd number >= 3 for HA)")
 	openbao.cmd.Flags().StringVar(&openbao.Opts.StorageSize, "storage-size", "10Gi", "PVC storage size for each OpenBao replica")
 	openbao.cmd.Flags().DurationVar(&openbao.Opts.Timeout, "timeout", 5*time.Minute, "Timeout for waiting on initialization")
+	openbao.cmd.Flags().StringVarP(&openbao.Opts.AgeKeyFile, "age-key-file", "k", "", "Path to age private key file for SOPS encryption/decryption (auto-detected if not set)")
+	openbao.cmd.Flags().BoolVarP(&openbao.Opts.Yes, "yes", "y", false, "Auto-approve fresh initialization when no DR backup is found")
 
 	util.MarkFlagRequired(openbao.cmd, "dr-backup-path")
 
--- a/cli/cmd/install_openbao.go
+++ b/cli/cmd/install_openbao.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"os/exec"
 	"os/signal"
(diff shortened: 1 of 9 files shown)


internal/installer/openbao.go
```diff
@@ -333,30 +387,124 @@ func (o *OpenBaoInstaller) ApplyVaultCR() error {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      openBaoUnsealSecretName,
+				Namespace: o.Config.Namespace,
+			},
+			Data: o.backupUnsealKeys,
+		}
+		_, err = secretsClient.Create(o.ctx, secret, metav1.CreateOptions{})
+		if err != nil && !k8serrors.IsAlreadyExists(err) {
+			return fmt.Errorf("creating unseal secret from backup: %w", err)
+		}
```
OliverTrautvetter, line 493: When Create fails with AlreadyExists, we return nil without updating the existing secret's data. But I think the secret can have wrong/empty data

internal/installer/openbao.go
```diff
@@ -410,48 +558,117 @@ func (o *OpenBaoInstaller) ExtractAndEncrypt() error {
...
-// To prevent the old bank-vaults sidecar from re-creating the secret via its
-// retry loop, the Vault CR is deleted first and we wait for all pods to exit
-// before removing the secret.
-func (o *OpenBaoInstaller) DeleteStaleUnsealKeys() error {
+// The cleanup sequence is:
+//  1. Delete the Vault CR (stops the bank-vaults sidecar retry loop)
+//  2. Wait for all vault pods to terminate
+//  3. Delete PVCs (removes stale Raft data that would confuse initialization)
+//  4. Delete the unseal-keys Secret
+func (o *OpenBaoInstaller) CleanStaleInstallState() error {
```
OliverTrautvetter, line 618: This function has not tests?

internal/installer/openbao.go
```diff
@@ -138,15 +172,21 @@ func (o *OpenBaoInstaller) Install(ctx context.Context) error {
...
 	if !o.drBackupExists {
-		err = o.Logger.Step("Removing stale unseal keys", o.DeleteStaleUnsealKeys)
+		err = o.Logger.Step("Cleaning stale install state", o.CleanStaleInstallState)
 		if err != nil {
-			return fmt.Errorf("failed to remove stale unseal keys: %w", err)
+			return fmt.Errorf("failed to clean stale install state: %w", err)
 		}
 	}
 
+	err = o.Logger.Step("Ensuring namespace exists", func() error {
```
OliverTrautvetter: So we clean before we ensure that it exists? Sounds slightly weird 🤔 maybe we should ensure first?

internal/installer/openbao.go
```diff
@@ -252,18 +271,53 @@ func (o *OpenBaoInstaller) GeneratePassword() error {
...
+	// If the namespace doesn't exist yet, there's certainly no release in it,
+	// so we skip the Helm query (which would fail on a non-existent namespace).
+	_, nsErr := o.Clientset.CoreV1().Namespaces().Get(o.ctx, o.Config.Namespace, metav1.GetOptions{})
+	if nsErr != nil && !k8serrors.IsNotFound(nsErr) {
+		return fmt.Errorf("checking namespace %s: %w", o.Config.Namespace, nsErr)
+	}
+	if nsErr == nil {
+		rel, err := o.Helm.FindRelease(o.Config.Namespace, cfg.ReleaseName)
+		if err != nil {
+			return err
```
OliverTrautvetter: we should wrap it like

return fmt.Errorf("finding release %s in namespace %s: %w", cfg.ReleaseName, o.Config.Namespace, err)

cli/cmd/install_openbao.go
```diff
@@ -69,6 +85,32 @@ func (c *InstallOpenBaoCmd) RunE(_ *cobra.Command, _ []string) error {
 		return fmt.Errorf("initializing openbao installer: %w", err)
 	}
 
+	inst.ConfirmFunc = func() error {
+		if c.Opts.Yes {
+			return nil
+		}
+
+		fmt.Printf("\nWARNING: No DR backup found at: %s\n", c.Opts.DRBackupPath)
```
OliverTrautvetter: We use log. Can we change all the fmt.Print stuff to log.Print

internal/installer/openbao.go
```diff
@@ -333,30 +387,124 @@ func (o *OpenBaoInstaller) ApplyVaultCR() error {
...
+		return fmt.Errorf("updating unseal secret from backup: %w", err)
+	}
+	return nil
+}
+
+// WaitForPodsReady polls until the expected number of vault pods (matching the
+// configured replica count) are in Running phase with all containers Ready.
+// This ensures scaling operations have fully completed before reporting success.
+func (o *OpenBaoInstaller) WaitForPodsReady() error {
+	selector := labels.SelectorFromSet(labels.Set{"vault_cr": "openbao"}).String()
```
OliverTrautvetter: We have many times "vault_cr", "openbao", labels.SelectorFromSet...

so we should create some const
```
const vaultCRLabelKey   = "vault_cr"
const vaultCRLabelValue = "openbao"

var vaultCRLabelSelector = labels.SelectorFromSet(labels.Set{vaultCRLabelKey: vaultCRLabelValue}).String()
```

internal/installer/openbao.go
```diff
@@ -252,18 +271,53 @@ func (o *OpenBaoInstaller) GeneratePassword() error {
 }
 
 // DeployBankVaultsOperator installs or upgrades the Bank-Vaults Operator Helm chart.
-// This is idempotent via UpgradeChart with InstallIfNotExist.
+//
+// The operator is cluster-scoped (it creates ClusterRoles, ClusterRoleBindings)
+// and watches Vault CRs across all namespaces. If the operator is already
+// installed in a different namespace, we skip re-deployment — one instance
+// is sufficient for the entire cluster.
 func (o *OpenBaoInstaller) DeployBankVaultsOperator() error {
```
OliverTrautvetter, line 291: Looks like the namespace is created by two different ways:

cfg.CreateNamespace: true in DeployBankVaultsOperator tells Helm to create the namespace on InstallChart.
ensureNamespace manually creates it via the Kubernetes API.

internal/installer/openbao.go
```diff
@@ -252,18 +271,53 @@ func (o *OpenBaoInstaller) GeneratePassword() error {
...
+	}
+	if nsErr == nil {
+		rel, err := o.Helm.FindRelease(o.Config.Namespace, cfg.ReleaseName)
+		if err != nil {
+			return err
+		}
+		if rel != nil {
+			// Release exists in target namespace — upgrade in place.
+			return o.Helm.UpgradeChart(o.ctx, cfg, UpgradeChartOptions{})
+		}
```
OliverTrautvetter: hard to test, maybe extract this to a function that can be tested alone like CheckForReleaseInTargetNamespace

internal/installer/openbao.go
```diff
@@ -252,18 +271,53 @@ func (o *OpenBaoInstaller) GeneratePassword() error {
...
+	// deployed cluster-wide (in another namespace) by looking for its ClusterRole.
+	_, err := o.Clientset.RbacV1().ClusterRoles().Get(o.ctx, "vault-operator", metav1.GetOptions{})
+	if err == nil {
+		// Operator already installed in another namespace — skip.
+		o.Logger.Logf("Bank-Vaults Operator already installed in the cluster, skipping deployment")
+		return nil
+	}
+	if !k8serrors.IsNotFound(err) {
+		return fmt.Errorf("checking for existing vault-operator ClusterRole: %w", err)
+	}
```
OliverTrautvetter: maybe also a function that can be tested alone

internal/installer/export_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+package installer
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Test helpers — exported only during `go test` so external test packages
```
OliverTrautvetter, line 12: Looks like at least backupUnsealKeys is missing and not tested

internal/installer/openbao.go
```diff
@@ -333,30 +387,124 @@ func (o *OpenBaoInstaller) ApplyVaultCR() error {
 
 // WaitForInitialization polls the openbao-unseal-keys Secret until it contains
 // unseal key data, indicating that Bank-Vaults has completed initialization.
+//
+// When a DR backup was loaded (backupUnsealKeys is set), the function ensures
+// the secret exists with the backup's unseal keys on every poll iteration. This
+// handles the case where the bank-vaults operator deletes or recreates the
+// secret during Vault CR reconciliation — we simply re-apply it until the
+// operator settles and the sidecar can successfully unseal.
 func (o *OpenBaoInstaller) WaitForInitialization() error {
```
OliverTrautvetter, line 434: no new tests for the new logic in "WaitForInitialization"

cli/cmd/install_openbao.go
```diff
@@ -69,6 +85,32 @@ func (c *InstallOpenBaoCmd) RunE(_ *cobra.Command, _ []string) error {
...
+		fmt.Println("If you intended to restore from a backup, verify --dr-backup-path is correct.")
+		fmt.Print("\nType 'yes' to continue: ")
+
+		reader := bufio.NewReader(os.Stdin)
+		input, err := reader.ReadString('\n')
+		if err != nil && !errors.Is(err, io.EOF) {
+			return fmt.Errorf("failed to read confirmation: %w", err)
+		}
+		if strings.TrimSpace(strings.ToLower(input)) != "yes" {
+			return fmt.Errorf("aborted: type 'yes' to continue or pass --yes")
```
OliverTrautvetter: Confusing message. If it is aborted, how can we continue. Aborted sounds like a big error, but it is just a typo / failed confirm

cli/cmd/install_openbao.go
```diff
@@ -29,19 +33,30 @@ type InstallOpenBaoCmd struct {
...
 func (c *InstallOpenBaoCmd) RunE(_ *cobra.Command, _ []string) error {
 	if err := validateOpenBaoPrereqs(); err != nil {
 		return err
 	}
 
+	// If --age-key-file is provided, set SOPS_AGE_KEY_FILE so ResolveAgeKey
+	// picks it up. Otherwise, fall back to the normal auto-discovery chain.
+	if c.Opts.AgeKeyFile != "" {
+		if err := os.Setenv("SOPS_AGE_KEY_FILE", c.Opts.AgeKeyFile); err != nil {
+			return fmt.Errorf("setting SOPS_AGE_KEY_FILE: %w", err)
```
OliverTrautvetter: changes global process state permanently.

We should pass the key file path through the OpenBaoInstallerConfig

internal/installer/openbao.go
```diff
@@ -410,48 +558,117 @@ func (o *OpenBaoInstaller) ExtractAndEncrypt() error {
...
 	if delErr != nil && !k8serrors.IsNotFound(delErr) {
 		return fmt.Errorf("deleting Vault CR: %w", delErr)
 	}
+	if delErr == nil {
+		cleaned = append(cleaned, "Vault CR")
+	}
 
 	if err := o.waitForVaultPodsGone(); err != nil {
 		return err
 	}
```
OliverTrautvetter, line 636: Why do we need to wait for a clear if Vault CR didn't exist.

We should only wait for pods when a Vault CR was actually deleted

OliverTrautvetter (changes requested):
Some comments to clean the code a bit

OliverTrautvetter (approved):
Thank you for the hard work on this PR 🚀

PR #433, opened 2026-05-20
feat(bootstrap): Add Telemetry settings to codesphere config

BBesrour:
Adds support for configuring telemetry export. 
Adds new command-line flags to enable sending traces and span metrics


internal/bootstrap/gcp/install_config.go
```diff
@@ -329,6 +319,15 @@ func (b *GCPBootstrapper) UpdateInstallConfig() error {
...
 	}
 
+	if b.Env.CentralOtelPassword != "" || b.Env.LocalTraceExport {
+		b.Env.InstallConfig.Codesphere.TelemetryExport = &files.TelemetryExport{
+			Endpoint:     b.Env.CentralOtelEndpoint,
+			RemoteExport: b.Env.CentralOtelPassword != "",
+			Traces:       b.Env.LocalTraceExport,
+			SpanMetrics:  b.Env.CentralOtelSpanMetrics,
+		}
+	}
```
joka134: Can we add some verification to give users some feedback, when these flags are partially configured but ignored by oms? Should be done atm in `gcp.go`.

Like password and LocalTraceExport was configured but no endpoint

joka134 (changes requested):
Looks good, just one thing

internal/bootstrap/gcp/install_config.go
BBesrour: We have default value in case export is enabled but no import specified. But added some check for creds and endpoint

joka134 (approved):

PR #434, opened 2026-05-20
fix: add eab key and config to the correct expected places

schrodit:
I think the current implementation didn't work with what the private cloud installer expects. See the pc docs: https://docs.codesphere.com/private-cloud/cluster-ingress-ca-options

Basically the acme config is in `codesphere` and not in the `certificates` config. The solver is then in the `certificates`


NautiluX (approved):

OliverTrautvetter (approved):
Looks good, hopefully that fixes the problem 👍

PR #438, opened 2026-05-22
feat(argo-migration): add a command to deploy all secrets of the prod.vault.yaml as secret to the target cluster

schrodit:
Required for the migration to argocd where secrets are consumed by external secret operator.


NautiluX (approved):
while we can do it as separate command, I think it could also be just something that happens when installing Codesphere.

PR #439, opened 2026-05-22
feat(argocd): secure password input and optional deploy-dc-config params

Jcing95:
## Summary
- OCI registry password is now read from `OMS_REGISTRY_PASSWORD` env var or prompted
  interactively (removes `--registry-password` flag)
- Git repo password is now read from `OMS_GIT_PASSWORD` env var; git repo credential
  is skipped when not set (removes `--git-password` flag)
- `--dc-id` is now optional: local cluster secret is skipped when not provided
- OCI registry URL is now configurable via `--registry-url` flag (defaults to
  `ghcr.io/codesphere-cloud/charts`)
- Removes the standalone `argocd-repo-secret` subcommand (functionality covered by
  `--deploy-dc-config`)
- Wraps underlying error in `NewArgoCD` failure path for better debugging
- Adds OCI password validation before applying the Helm registry secret

## Usage
```bash
# Install ArgoCD chart only
oms beta install argocd

# Install chart + apply Codesphere resources (prompts for OCI password)
oms beta install argocd --deploy-dc-config

# Full install with all resources
OMS_REGISTRY_PASSWORD=token OMS_GIT_PASSWORD=git-token oms beta install argocd --deploy-dc-config --dc-id 0

# Use a custom/mirrored OCI registry
OMS_REGISTRY_PASSWORD=token oms beta install argocd --deploy-dc-config --registry-url my-registry.example.com/charts
```

## Breaking Changes
- `--registry-password` flag removed (use `OMS_REGISTRY_PASSWORD` env var instead)
- `--git-password` flag removed (use `OMS_GIT_PASSWORD` env var instead)
- `oms beta install argocd-repo-secret` subcommand removed


internal/installer/argocd_resources.go
```diff
@@ -60,16 +60,24 @@ func (a *argoCDResources) ApplyAll(ctx context.Context) error {
...
-	if err := a.applyLocalCluster(ctx); err != nil {
-		return fmt.Errorf("applying local cluster secret: %w", err)
+	if a.DatacenterId != "" {
+		if err := a.applyLocalCluster(ctx); err != nil {
+			return fmt.Errorf("applying local cluster secret: %w", err)
+		}
+	}
+
+	if a.OciPassword == "" {
+		return fmt.Errorf("OCI registry password is required but not set")
```
schrodit: We do not template so no need for a templated error
```suggestion
		return errors.New("OCI registry password is required but not set")
```

internal/installer/argocd_resources.go
```diff
@@ -60,16 +60,24 @@ func (a *argoCDResources) ApplyAll(ctx context.Context) error {
 		return fmt.Errorf("applying app projects: %w", err)
```
schrodit, line 63: I'm also wondering if we really need the projects here: https://github.com/codesphere-cloud/oms/blob/main/internal/installer/manifests/argocd/app-projects.yaml

I think just the default project is totally fine for us. Also we do not really have dev/prod in a single cluster.

But adding @bachgg since you added it

cli/cmd/argocd.go
```diff
@@ -18,30 +20,26 @@ type InstallArgoCDCmd struct {
...
 		}
+		ociPassword = pw
+		gitPassword = os.Getenv("OMS_GIT_PASSWORD")
 	}
-	install, err := installer.NewArgoCD(c.Opts.Version, c.Opts.DatacenterId, c.Opts.RegistryPassword, c.Opts.GitPassword, c.Opts.FullInstall)
+
+	install, err := installer.NewArgoCD(c.Opts.Version, c.Opts.DatacenterId, ociPassword, gitPassword, c.Opts.FullInstall)
 	if err != nil {
-		return fmt.Errorf("failed to initialize ArgoCD installer")
+		return fmt.Errorf("failed to initialize ArgoCD installer: %w", err)
```
schrodit, line 47: thanks a did the exact same change:)

internal/installer/argocd_resources.go
```diff
@@ -60,16 +61,24 @@ func (a *argoCDResources) ApplyAll(ctx context.Context) error {
...
 	}
 
-	if err := a.applyLocalCluster(ctx); err != nil {
-		return fmt.Errorf("applying local cluster secret: %w", err)
+	if a.DatacenterId != "" {
+		if err := a.applyLocalCluster(ctx); err != nil {
```
schrodit, line 67: same question as above, I don't think we really need those resources. but lets double check with @bachgg

internal/installer/argocd_resources.go
```diff
@@ -60,16 +61,24 @@ func (a *argoCDResources) ApplyAll(ctx context.Context) error {
...
+			return fmt.Errorf("applying local cluster secret: %w", err)
+		}
+	}
+
+	if a.OciPassword == "" {
+		return errors.New("OCI registry password is required but not set")
 	}
 
 	if err := a.applyHelmRegistrySecret(ctx); err != nil {
```
schrodit, line 76: the used oci registry is currently hardcoded. maybe you can also make it configurable

schrodit (commented):
mostly fine just some clarification missing

internal/installer/argocd_resources.go
bachgg: The dev/prod projects are meant for argocd instances that read from the `charts` repo and where different people get access to it. In single-instance clusters, the default project is okay.

internal/installer/argocd_resources.go
bachgg: This is for reading values.yaml files from `charts` repo, can also be removed now that values are deployed together with applications

bachgg (approved):
General logic change looks good to me

schrodit (approved):

OliverTrautvetter (approved):

PR #442, opened 2026-05-22
feat(argocd): add install pc-apps command

Jcing95:
## Summary

Adds a new `oms beta install pc-apps` CLI command that installs/upgrades the pc-applications Helm chart from a private OCI registry. The chart deploys ArgoCD Application resources that manage platform components.

## Key Design Decisions

- **Credentials from K8s Secret** — registry URL and credentials are read from the K8s Secret `argocd-codesphere-oci-read` (created by `oms beta install argocd --deploy-dc-config`). No manual auth flags needed.
- **Eager registry login** — `LoginRegistry` performs the real OCI login immediately so bad credentials fail fast with a clear error
- **Values file merging** — multiple `-f` files are merged in order using Helm's built-in values merge, allowing base + overlay patterns

## Usage

```bash
# Install a specific version (version is required)
oms beta install pc-apps --version 1.2.3

# Install with custom values files
oms beta install pc-apps --version 1.0.0 -f base.yaml -f dc-overlay.yaml

# Install into a custom namespace
oms beta install pc-apps --version 1.0.0 --namespace custom-ns
```

## Changes

- `cli/cmd/pc_apps.go` — Cobra command with flags: `--version` (required), `--namespace`, `--values`/`-f`
- `cli/cmd/beta_install.go` — Registers pc-apps under `oms beta install`
- `internal/installer/pc_apps.go` — Install/upgrade logic reading credentials from K8s Secret and merging values files
- `internal/installer/helm_client.go` — Adds `LoginRegistry` to `HelmClient` interface; performs eager OCI login and reuses the authenticated registry client
- `internal/installer/pc_apps_test.go` — Tests covering install, registry login failure, missing secret, missing fields, and values merging
- `internal/installer/mocks.go` — Updated generated mocks for `HelmClient`

### DIFF
--- a/cli/cmd/pc_apps.go
+++ b/cli/cmd/pc_apps.go
@@ -9,8 +9,6 @@ import (
 	"github.com/codesphere-cloud/oms/internal/installer"
 	"github.com/codesphere-cloud/oms/internal/util"
 	"github.com/spf13/cobra"
-	"k8s.io/apimachinery/pkg/runtime"
-	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlconfig "sigs.k8s.io/controller-runtime/pkg/client/config"
 )
@@ -34,12 +32,7 @@ func (c *InstallPCAppsCmd) RunE(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to load kubernetes config: %w", err)
 	}
 
-	scheme := runtime.NewScheme()
-	if err := clientgoscheme.AddToScheme(scheme); err != nil {
-		return fmt.Errorf("failed to add kubernetes scheme: %w", err)
-	}
-
-	kubeClient, err := ctrlclient.New(kubeConfig, ctrlclient.Options{Scheme: scheme})
+	kubeClient, err := ctrlclient.New(kubeConfig, ctrlclient.Options{})
 	if err != nil {
 		return fmt.Errorf("failed to create kubernetes client: %w", err)
 	}
--- a/cli/cmd/pc_apps.go
+++ b/cli/cmd/pc_apps.go
@@ -4,12 +4,15 @@ package cmd
 
 import (
 	"fmt"
-	"os"
 
 	packageio "github.com/codesphere-cloud/cs-go/pkg/io"
 	"github.com/codesphere-cloud/oms/internal/installer"
+	"github.com/codesphere-cloud/oms/internal/util"
 	"github.com/spf13/cobra"
-	"golang.org/x/term"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlconfig "sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 // InstallPCAppsCmd represents the pc-apps command
@@ -20,29 +23,31 @@ type InstallPCAppsCmd struct {
 
 type InstallPCAppsOpts struct {
 	*GlobalOptions
-	Chart       string
 	Version     string
 	Namespace   string
-	Username    string
 	ValuesFiles []string
 }
 
 func (c *InstallPCAppsCmd) RunE(cmd *cobra.Command, args []string) error {
-	var password string
-	if c.Opts.Username != "" {
-		pw, err := c.resolvePassword()
-		if err != nil {
-			return err
-		}
-		password = pw
+	kubeConfig, err := ctrlconfig.GetConfig()
+	if err != nil {
+		return fmt.Errorf("failed to load kubernetes config: %w", err)
+	}
+
+	scheme := runtime.NewScheme()
+	if err := clientgoscheme.AddToScheme(scheme); err != nil {
+		return fmt.Errorf("failed to add kubernetes scheme: %w", err)
+	}
+
+	kubeClient, err := ctrlclient.New(kubeConfig, ctrlclient.Options{Scheme: scheme})
+	if err != nil {
+		return fmt.Errorf("failed to create kubernetes client: %w", err)
 	}
 
 	pcApps, err := installer.NewPCApps(
-		c.Opts.Chart,
+		kubeClient,
 		c.Opts.Version,
 		c.Opts.Namespace,
-		c.Opts.Username,
-		password,
 		c.Opts.ValuesFiles,
 	)
 	if err != nil {
@@ -56,25 +61,6 @@ func (c *InstallPCAppsCmd) RunE(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-// resolvePassword reads the password from the OMS_REPO_PASSWORD environment variable,
-// or prompts the user interactively if the env var is not set.
-func (c *InstallPCAppsCmd) resolvePassword() (string, error) {
-	if pw := os.Getenv("OMS_REPO_PASSWORD"); len(pw) != 0 {
-		return pw, nil
-	}
-
-	fmt.Print("Registry password/token: ")
-	pw, err := term.ReadPassword(int(os.Stdin.Fd()))
-	fmt.Println()
-	if err != nil {
-		return "", fmt.Errorf("failed to read password: %w", err)
-	}
-	if len(pw) == 0 {
-		return "", fmt.Errorf("password is required; set OMS_REPO_PASSWORD or enter it when prompted")
-	}
-	return string(pw), nil
-}
-
 func AddPCAppsCmd(parentCmd *cobra.Command, opts *GlobalOptions) {
 	pcApps := InstallPCAppsCmd{
 		Opts: InstallPCAppsOpts{
@@ -83,28 +69,26 @@ func AddPCAppsCmd(parentCmd *cobra.Command, opts *GlobalOptions) {
 	}
 	pcApps.cmd = &cobra.Command{
 		Use:   "pc-apps",
-		Short: "Install the pc-apps Helm chart from a private OCI registry",
-		Long: packageio.Long(`Install or upgrade the pc-apps Helm chart from a private OCI registry
-			into the target cluster. This chart deploys ArgoCD Application resources
-			that manage the platform components.
+		Short: "Install the pc-applications Helm chart from a private OCI registry",
+		Long: packageio.Long(`Install or upgrade the pc-applications Helm chart from a private OCI
+			registry into the target cluster. This chart deploys ArgoCD Application
+			resources that manage the platform components.
 
-			If --username is provided, the registry password is read from the
-			OMS_REPO_PASSWORD environment variable or prompted interactively.
-			Otherwise, credentials are read from the Kubernetes secret
-			"argocd-codesphere-oci-read" in the argocd namespace (created by
-			"oms beta install argocd --deploy-dc-config --registry-password <token>").`),
+			Registry credentials and chart URL are read automatically from the
+			Kubernetes secret "argocd-codesphere-oci-read" in the argocd namespace.
+			This secret is created by "oms beta install argocd --deploy-dc-config".`),
 		Example: formatExamples("beta install pc-apps", []packageio.Example{
-			{Cmd: "--version 1.0.0", Desc: "Install a specific version (credentials from K8s secret)"},
-			{Cmd: "--version 1.0.0 --username CodesphereBot", Desc: "Install with explicit registry credentials (prompts for password)"},
-			{Cmd: "--chart oci://ghcr.io/codesphere-cloud/charts/pc-apps --version 1.0.0 -f base.yaml -f dc-overlay.yaml", Desc: "Install with custom chart and values files"},
+			{Cmd: "--version 1.0.0", Desc: "Install a specific version"},
+			{Cmd: "--version 1.0.0 -f base.yaml -f dc-overlay.yaml", Desc: "Install with custom values files"},
+			{Cmd: "--version 1.0.0 --namespace custom-ns", Desc: "Install into a custom namespace"},
 		}),
 	}
-	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Chart, "chart", "oci://ghcr.io/codesphere-cloud/charts/pc-apps", "Full OCI chart URL")
-	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Version, "version", "", "Chart version to install (default: latest)")
+	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Version, "version", "", "Chart version to install (required)")
 	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Namespace, "namespace", "argocd", "Target namespace for the Helm release")
-	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Username, "username", "", "Username for OCI registry authentication (if omitted, reads from K8s secret)")
 	pcApps.cmd.Flags().StringArrayVarP(&pcApps.Opts.ValuesFiles, "values", "f", nil, "Path to values YAML file (can be specified multiple times, merged in order)")
 	pcApps.cmd.RunE = pcApps.RunE
+
+	util.MarkFlagRequired(pcApps.cmd, "version")
 
 	AddCmd(parentCmd, pcApps.cmd)
 }
--- a/cli/cmd/pc_apps.go
+++ b/cli/cmd/pc_apps.go
@@ -3,7 +3,6 @@
 package cmd
 
 import (
-	"context"
 	"fmt"
 	"os"
 
@@ -28,7 +27,7 @@ type InstallPCAppsOpts struct {
 	ValuesFiles []string
 }
 
-func (c *InstallPCAppsCmd) RunE(_ *cobra.Command, args []string) error {
+func (c *InstallPCAppsCmd) RunE(cmd *cobra.Command, args []string) error {
 	var password string
 	if c.Opts.Username != "" {
 		pw, err := c.resolvePassword()
@@ -50,7 +49,7 @@ func (c *InstallPCAppsCmd) RunE(_ *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to initialize pc-apps installer: %w", err)
 	}
 
-	if err := pcApps.Install(context.Background()); err != nil {
+	if err := pcApps.Install(cmd.Context()); err != nil {
 		return fmt.Errorf("failed to install pc-apps: %w", err)
 	}
 
@@ -93,7 +92,7 @@ func AddPCAppsCmd(parentCmd *cobra.Command, opts *GlobalOptions) {
 			OMS_REPO_PASSWORD environment variable or prompted interactively.
 			Otherwise, credentials are read from the Kubernetes secret
 			"argocd-codesphere-oci-read" in the argocd namespace (created by
-			"oms beta install argocd").`),
+			"oms beta install argocd --deploy-dc-config --registry-password <token>").`),
 		Example: formatExamples("beta install pc-apps", []packageio.Example{
 			{Cmd: "--version 1.0.0", Desc: "Install a specific version (credentials from K8s secret)"},
 			{Cmd: "--version 1.0.0 --username CodesphereBot", Desc: "Install with explicit registry credentials (prompts for password)"},
--- a/cli/cmd/pc_apps.go
+++ b/cli/cmd/pc_apps.go
@@ -29,20 +29,13 @@ type InstallPCAppsOpts struct {
 }
 
 func (c *InstallPCAppsCmd) RunE(_ *cobra.Command, args []string) error {
-	requiredFlags := map[string]string{
-		"chart":    c.Opts.Chart,
-		"username": c.Opts.Username,
-	}
-
-	for flagName, value := range requiredFlags {
-		if value == "" {
-			return fmt.Errorf("flag --%s is required", flagName)
+	var password string
+	if c.Opts.Username != "" {
+		pw, err := c.resolvePassword()
+		if err != nil {
+			return err
 		}
-	}
-
-	password, err := c.resolvePassword()
-	if err != nil {
-		return err
+		password = pw
 	}
 
 	pcApps, err := installer.NewPCApps(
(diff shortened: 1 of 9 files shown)


cli/cmd/pc_apps.go
```diff
@@ -0,0 +1,111 @@
...
+		return fmt.Errorf("failed to install pc-apps: %w", err)
+	}
+
+	return nil
+}
+
+// resolvePassword reads the password from the OMS_REPO_PASSWORD environment variable,
+// or prompts the user interactively if the env var is not set.
+func (c *InstallPCAppsCmd) resolvePassword() (string, error) {
+	if pw := os.Getenv("OMS_REPO_PASSWORD"); len(pw) != 0 {
```
schrodit: maybe better call it `OMS_REGISTRY_PASSWORD`?

cli/cmd/pc_apps.go
```diff
@@ -0,0 +1,111 @@
...
+			"argocd-codesphere-oci-read" in the argocd namespace (created by
+			"oms beta install argocd").`),
+		Example: formatExamples("beta install pc-apps", []packageio.Example{
+			{Cmd: "--version 1.0.0", Desc: "Install a specific version (credentials from K8s secret)"},
+			{Cmd: "--version 1.0.0 --username CodesphereBot", Desc: "Install with explicit registry credentials (prompts for password)"},
+			{Cmd: "--chart oci://ghcr.io/codesphere-cloud/charts/pc-apps --version 1.0.0 -f base.yaml -f dc-overlay.yaml", Desc: "Install with custom chart and values files"},
+		}),
+	}
+	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Chart, "chart", "oci://ghcr.io/codesphere-cloud/charts/pc-apps", "Full OCI chart URL")
+	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Version, "version", "", "Chart version to install (default: latest)")
```
schrodit: there is not default configured currently.
But I also don't think we should default to latest (we also currently do not publish a latest version).
Lets rather stay explicit and always require a version.
Later the version will anyway be part of the bundle that is installed, so no manual intervention required

cli/cmd/pc_apps.go
```diff
@@ -0,0 +1,111 @@
...
+			Otherwise, credentials are read from the Kubernetes secret
+			"argocd-codesphere-oci-read" in the argocd namespace (created by
+			"oms beta install argocd").`),
+		Example: formatExamples("beta install pc-apps", []packageio.Example{
+			{Cmd: "--version 1.0.0", Desc: "Install a specific version (credentials from K8s secret)"},
+			{Cmd: "--version 1.0.0 --username CodesphereBot", Desc: "Install with explicit registry credentials (prompts for password)"},
+			{Cmd: "--chart oci://ghcr.io/codesphere-cloud/charts/pc-apps --version 1.0.0 -f base.yaml -f dc-overlay.yaml", Desc: "Install with custom chart and values files"},
+		}),
+	}
+	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Chart, "chart", "oci://ghcr.io/codesphere-cloud/charts/pc-apps", "Full OCI chart URL")
```
schrodit:
```suggestion
	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Chart, "chart", "oci://ghcr.io/codesphere-cloud/charts/pc-applications", "OCI chart URL")
```

cli/cmd/pc_apps.go
```diff
@@ -0,0 +1,111 @@
...
+	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Version, "version", "", "Chart version to install (default: latest)")
+	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Namespace, "namespace", "argocd", "Target namespace for the Helm release")
+	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Username, "username", "", "Username for OCI registry authentication (if omitted, reads from K8s secret)")
```
schrodit: I'm wondering if we need all of that when we anyway read from the secret?
I think we will anyway integrate it somehow directly, then we can still think about a nicer UX.

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,238 @@
...
+// a private OCI registry.
+type PCApps struct {
+	ChartURL    string   // full OCI chart reference, e.g. "oci://ghcr.io/codesphere-cloud/charts/pc-apps"
+	Version     string   // chart version ("" means latest)
+	Namespace   string   // target namespace (default: "argocd")
+	Username    string   // OCI registry username (optional, falls back to K8s secret)
+	Password    string   // OCI registry password/token (optional, falls back to K8s secret)
+	ValuesFiles []string // paths to values YAML files, merged in order
+	Helm        HelmClient
+	Clientset   kubernetes.Interface
```
schrodit: rather use the `client.Client` package and not the raw k8s interface. (its simpler to use and also used across the repo)

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// NewPCApps creates a new PCApps installer with a real Helm client and
+// Kubernetes clientset for credential fallback.
+func NewPCApps(chartURL, version, namespace, username, password string, valuesFiles []string) (*PCApps, error) {
+	if namespace == "" {
+		namespace = "argocd"
```
schrodit: the defaulting should rather happen on cli level. here we should only validate.
Otherwise we have defaults in multi places which makes it hard to read

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	existing, err := p.Helm.FindRelease(p.Namespace, releaseName)
+	if err != nil {
+		return fmt.Errorf("checking existing release: %w", err)
+	}
+
+	if existing != nil {
+		log.Printf("Found existing release %s (chart version %s), upgrading...\n", releaseName, existing.InstalledVersion)
+		if err := p.Helm.UpgradeChart(ctx, cfg, UpgradeChartOptions{}); err != nil {
```
schrodit: why not use the internal `InstallIfNotExist` instead of handling the create/update yourself?


```suggestion
		if err := p.Helm.UpgradeChart(ctx, cfg, UpgradeChartOptions{InstallIfNotExist: true}); err != nil {
```

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,238 @@
...
+		srcMap, srcOk := srcVal.(map[string]interface{})
+		dstMap, dstOk := dstVal.(map[string]interface{})
+		if srcOk && dstOk {
+			dst[key] = deepMerge(dstMap, srcMap)
+		} else {
+			dst[key] = srcVal
+		}
+	}
+	return dst
+}
```
schrodit: we shouldn't do that on our own. lets rather use the upstream logic from helm directly: e.g. https://github.com/helm/helm/blob/main/pkg/cli/values/options.go#L45

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,238 @@
...
+func parseOCIChartURL(chartURL string) (host string, chartName string, err error) {
+	if !strings.HasPrefix(chartURL, "oci://") {
+		return "", "", fmt.Errorf("chart URL must start with \"oci://\", got %q", chartURL)
+	}
+
+	// Replace oci:// with https:// for standard URL parsing
+	parsed, err := url.Parse(strings.Replace(chartURL, "oci://", "https://", 1))
+	if err != nil {
+		return "", "", fmt.Errorf("parsing chart URL %q: %w", chartURL, err)
+	}
```
schrodit: don't think the replace is required. `url.Parse` can handle `oci` see https://go.dev/play/p/FK-XbPt4YCR

```suggestion
	parsed, err := url.Parse(chartURL))
	if err != nil {
		return "", "", fmt.Errorf("parsing chart URL %q: %w", chartURL, err)
	}
```

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,238 @@
...
+
+	log.Printf("Using credentials from K8s secret %q\n", ociCredentialSecretName)
+	return username, password, nil
+}
+
+// Install authenticates against the OCI registry and installs or upgrades the
+// pc-apps Helm chart.
+func (p *PCApps) Install(ctx context.Context) error {
+	host, releaseName, err := parseOCIChartURL(p.ChartURL)
```
schrodit: I would not set the release name baed on the helm chart.
lets rather hard-code it. This command is very specific to our set of apps so its fine that we are that specific now.

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,238 @@
...
+	// Replace oci:// with https:// for standard URL parsing
+	parsed, err := url.Parse(strings.Replace(chartURL, "oci://", "https://", 1))
+	if err != nil {
+		return "", "", fmt.Errorf("parsing chart URL %q: %w", chartURL, err)
+	}
+
+	host = parsed.Host
+	if host == "" {
+		return "", "", fmt.Errorf("chart URL %q has no host", chartURL)
+	}
```
schrodit: whats the reason for another variable?

```suggestion
	if parsed.Host == "" {
		return "", "", fmt.Errorf("chart URL %q has no host", chartURL)
	}
```

cli/cmd/pc_apps.go
```diff
@@ -0,0 +1,94 @@
...
+func (c *InstallPCAppsCmd) RunE(cmd *cobra.Command, args []string) error {
+	kubeConfig, err := ctrlconfig.GetConfig()
+	if err != nil {
+		return fmt.Errorf("failed to load kubernetes config: %w", err)
+	}
+
+	scheme := runtime.NewScheme()
+	if err := clientgoscheme.AddToScheme(scheme); err != nil {
+		return fmt.Errorf("failed to add kubernetes scheme: %w", err)
+	}
```
schrodit: isn't the runtime scheme already included by default?

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+// PCApps holds the configuration for installing the pc-applications Helm chart
+// from a private OCI registry.
+type PCApps struct {
+	Version     string   // chart version (required)
+	Namespace   string   // target namespace for the Helm release
+	ValuesFiles []string // paths to values YAML files, merged in order
+	Helm        HelmClient
+	Client      client.Client
```
schrodit: nit: no need to have any of those public?

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// NewPCApps creates a new PCApps installer. It validates that required fields
+// are non-empty but does not apply defaults — defaults live on the CLI flag
+// declarations only.
+func NewPCApps(c client.Client, version, namespace string, valuesFiles []string) (*PCApps, error) {
+	if version == "" {
+		return nil, fmt.Errorf("version is required")
```
schrodit:
```suggestion
		return nil, errors.New("version is required")
```

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	if namespace == "" {
+		return nil, fmt.Errorf("namespace is required")
```
schrodit:
```suggestion
		return nil, errors.New("namespace is required")
```

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,139 @@
...
+	password = string(secret.Data["password"])
+
+	if baseURL == "" || username == "" || password == "" {
+		return "", "", "", fmt.Errorf(
+			"K8s secret %q in namespace %q is missing required fields (url, username, or password)",
+			ociCredentialSecretName, ociCredentialNamespace,
+		)
+	}
+
+	chartURL = "oci://" + baseURL + "/" + pcAppsReleaseName
```
schrodit: the releaseName is something different than the chartName.
Lets have 2 different const for that

internal/installer/pc_apps_test.go
```diff
@@ -0,0 +1,264 @@
...
+				Build()
+			pcApps = &installer.PCApps{
+				Version:   version,
+				Namespace: namespace,
+				Helm:      helmMock,
+				Client:    fakeClient,
+			}
+		})
+
+		It("returns a clear error suggesting how to fix", func() {
```
schrodit: not sure if we really return an error how to fix it.

```suggestion
		It("returns a clear error", func() {
```

schrodit (approved):
I think in the laster version we should not get the chartUrl only from the secret, but rather give it from the outside, but thats fine for now

internal/installer/pc_apps.go
Jcing95: They have to be public for the tests to be able to construct this as the tests are in a different package

cli/cmd/pc_apps.go
Jcing95: The core types apparently are not autoregistered so we need a explicit scheme here. :thinking: 
(according to claude)

internal/installer/pc_apps.go
schrodit: usually you would rather have a dedicated `NewForTesting` that only exposes the required fields for testing.

cli/cmd/pc_apps.go
schrodit: according to the examples int eh repo its included: e.g. https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#example-New

see that code that k8s scheme is the default: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/client.go#L161

schrodit: ping

internal/installer/pc_apps.go
schrodit: ping

schrodit (approved):
lgtm

internal/installer/pc_apps_test.go
```diff
@@ -0,0 +1,226 @@
...
+			helmMock.EXPECT().LoginRegistry(mock.Anything, "ghcr.io", secretUsername, secretPassword).Return(nil)
+			helmMock.EXPECT().UpgradeChart(mock.Anything, mock.MatchedBy(func(cfg installer.ChartConfig) bool {
+				nested, ok := cfg.Values["nested"].(map[string]any)
+				if !ok {
+					return false
+				}
+				return cfg.Values["foo"] == "overridden" &&
+					fmt.Sprint(nested["a"]) == "1" &&
+					fmt.Sprint(nested["b"]) == "99" &&
+					fmt.Sprint(nested["c"]) == "3"
```
OliverTrautvetter: Nice trick but we should just parse it. 
`nested["a"].(int) == 1`

internal/installer/pc_apps_test.go
```diff
@@ -0,0 +1,226 @@
...
+		})
+
+		It("reads credentials from K8s secret and calls UpgradeChart with InstallIfNotExist", func() {
+			helmMock.EXPECT().LoginRegistry(mock.Anything, "ghcr.io", secretUsername, secretPassword).Return(nil)
+			helmMock.EXPECT().UpgradeChart(mock.Anything, mock.MatchedBy(func(cfg installer.ChartConfig) bool {
+				return cfg.ReleaseName == "pc-applications" &&
+					cfg.ChartName == expectedChartURL &&
+					cfg.Namespace == namespace &&
+					cfg.Version == version &&
+					cfg.CreateNamespace == true
```
OliverTrautvetter: nit, redundant comparison, just `cfg.CreateNamespace` is enough
we don't need to validate a bool with true

cli/cmd/pc_apps.go
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Codesphere Inc. SPDX-License-Identifier: Apache-2.0
```
OliverTrautvetter:
```suggestion
// Copyright (c) Codesphere Inc.
// SPDX-License-Identifier: Apache-2.0
```

OliverTrautvetter: oh, please also change it like this in `argocd.go` to keep everything consistent 😄

cli/cmd/pc_apps.go
```diff
@@ -0,0 +1,87 @@
...
+			Registry credentials and chart URL are read automatically from the
+			Kubernetes secret "argocd-codesphere-oci-read" in the argocd namespace.
+			This secret is created by "oms beta install argocd --deploy-dc-config".`),
+		Example: formatExamples("beta install pc-apps", []packageio.Example{
+			{Cmd: "--version 1.0.0", Desc: "Install a specific version"},
+			{Cmd: "--version 1.0.0 -f base.yaml -f dc-overlay.yaml", Desc: "Install with custom values files"},
+			{Cmd: "--version 1.0.0 --namespace custom-ns", Desc: "Install into a custom namespace"},
+		}),
+	}
+	pcApps.cmd.Flags().StringVar(&pcApps.Opts.Version, "version", "", "Chart version to install (required)")
```
OliverTrautvetter, line 80: pr desc says 

"version to latest" and usage examples show --username and --chart flags that don't exist in the code

So desc is wrong?

internal/installer/helm_client.go
```diff
@@ -73,6 +77,20 @@ func NewHelmClient(namespace string) (HelmClient, error) {
 	}, nil
 }
 
+func (h *helmClient) LoginRegistry(_ context.Context, host, username, password string) error {
```
OliverTrautvetter, line 83: context parameter is silently ignored, maybe add a comment

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,154 @@
...
+const (
+	// ociCredentialSecretName is the K8s Secret created by "oms beta install argocd"
+	// that stores OCI registry credentials for pulling Codesphere charts.
+	ociCredentialSecretName = "argocd-codesphere-oci-read"
+	// ociCredentialNamespace is the namespace where the credential secret lives.
+	ociCredentialNamespace = "argocd"
+	// pcAppsReleaseName is the fixed Helm release name for the pc-applications chart.
+	pcAppsReleaseName = "pc-applications"
+	// pcAppsChartName is the chart name used when constructing the OCI chart URL.
+	pcAppsChartName = "pc-applications"
```
OliverTrautvetter, line 30: 2 const with the same value, why? Just use 1?

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,154 @@
...
+func (p *PCApps) resolveFromSecret(ctx context.Context) (chartURL, username, password string, err error) {
+	secret := &corev1.Secret{}
+	key := client.ObjectKey{Name: ociCredentialSecretName, Namespace: ociCredentialNamespace}
+	if err := p.client.Get(ctx, key, secret); err != nil {
+		return "", "", "", fmt.Errorf(
+			"K8s secret %q not found in namespace %q: %w\n"+
+				"Run 'oms beta install argocd --deploy-dc-config' first to create registry credentials",
+			ociCredentialSecretName, ociCredentialNamespace, err,
+		)
+	}
```
OliverTrautvetter: The named return err is declared but never used. The inner `if err :=` shadows it throughout the function. 

So either remove the named return or remove the shadowing :=

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,154 @@
...
+	password = string(secret.Data["password"])
+
+	if baseURL == "" || username == "" || password == "" {
+		return "", "", "", fmt.Errorf(
+			"K8s secret %q in namespace %q is missing required fields (url, username, or password)",
+			ociCredentialSecretName, ociCredentialNamespace,
+		)
+	}
+
+	chartURL = "oci://" + baseURL + "/" + pcAppsChartName
```
OliverTrautvetter: use url.JoinPath to be safe

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,154 @@
...
+		Values:          vals,
+		CreateNamespace: true,
+	}
+
+	log.Printf("Installing/Upgrading %s (version %s) into namespace %s\n", pcAppsReleaseName, p.version, p.namespace)
+	if err := p.helm.UpgradeChart(ctx, cfg, UpgradeChartOptions{InstallIfNotExist: true}); err != nil {
+		return fmt.Errorf("install/upgrade failed: %w", err)
+	}
+
+	fmt.Printf("Successfully installed/upgraded %s\n", pcAppsReleaseName)
```
OliverTrautvetter:
```suggestion
	log.Printf("Successfully installed/upgraded %s\n", pcAppsReleaseName)
```

internal/installer/pc_apps.go
```diff
@@ -0,0 +1,154 @@
...
+	return nil
+}
+
+// NewPCAppsForTesting creates a PCApps instance with injected dependencies for
+// use in tests. This avoids exporting struct fields solely for test access.
+func NewPCAppsForTesting(helm HelmClient, c client.Client, version, namespace string, valuesFiles []string) *PCApps {
```
OliverTrautvetter, line 150: This function should not be here but in the pc_apps_test.go file

OliverTrautvetter (changes requested):
Some comments to keep the code clean

internal/installer/pc_apps.go
Jcing95: they are tecnically different things, and the value of one may change in the future. Did two since Tim requested it :)

internal/installer/pc_apps.go
Jcing95: Since the fields are not exported the constructor has to be in the same package. Either new file or here. Alternative would be to make the fields public again. WDYT @schrodit @OliverTrautvetter?

OliverTrautvetter: You are right, I think it is fine in that case

internal/installer/pc_apps_test.go
```diff
@@ -195,9 +194,9 @@ var _ = Describe("PCApps.Install", func() {
 					return false
 				}
 				return cfg.Values["foo"] == "overridden" &&
-					fmt.Sprint(nested["a"]) == "1" &&
-					fmt.Sprint(nested["b"]) == "99" &&
-					fmt.Sprint(nested["c"]) == "3"
+					nested["a"].(float64) == 1 &&
```
OliverTrautvetter, line 197: why float64 and not int?

OliverTrautvetter (approved):
Approved, just one question

internal/installer/pc_apps_test.go
Jcing95: This is what claude says for that matter :D

> The values go through sigs.k8s.io/yaml which internally converts YAML→JSON→Go, and encoding/json always produces float64 for numbers in interface{} targets. Using .(int) as Oliver suggested would cause a runtime panic

OliverTrautvetter: 😅 good point... yeah that is weird. But looks good in that case

PR #453, opened 2026-05-27
feat(bootstrap): add Prometheus remote write configuration options

BBesrour:
Adds support for configuring Prometheus remote write in the GCP bootstrap process.

Adds flags for specifying the Prometheus remote write config + validation

Writes prom credentials in the vault when passed


internal/installer/files/config_yaml.go
```diff
@@ -851,6 +854,24 @@ func (c *RootConfig) addMonitoringSecrets(vault *InstallVault) {
...
 	}
+
+	if c.Cluster.Monitoring != nil && c.Cluster.Monitoring.Prometheus != nil && c.Cluster.Monitoring.Prometheus.RemoteWrite != nil {
+		if c.Cluster.Monitoring.Prometheus.RemoteWrite.Username != "" && c.Cluster.Monitoring.Prometheus.RemoteWrite.Password != "" {
+			vault.SetSecret(SecretEntry{
+				Name: "promRemoteWriteUser",
+				Fields: &SecretFields{
+					Password: c.Cluster.Monitoring.Prometheus.RemoteWrite.Username,
+				},
+			})
```
OliverTrautvetter, line 865:
```suggestion
			vault.SetSecret(SecretEntry{
				Name: "promRemoteWriteUser",
				Fields: &SecretFields{
					Username: c.Cluster.Monitoring.Prometheus.RemoteWrite.Username,
				},
			})
```

OliverTrautvetter: Username should be Username. 

Probably even better would be something like we do for `centralOtelCreds`

```
vault.SetSecret(SecretEntry{
				Name: "name",
				Fields: &SecretFields{
					Username: xy
					Password: xy,
				},
			})
```

instead of 2 separate secrets. Or maybe there is a reason?

internal/bootstrap/gcp/gcp.go
```diff
@@ -522,6 +530,16 @@ func (b *GCPBootstrapper) validateExternalLokiParams() error {
 	return nil
 }
 
+func (b *GCPBootstrapper) validatePrometheusRemoteWriteParams() error {
+	if b.Env.PrometheusRemoteWriteURL != "" && (b.Env.PrometheusRemoteWriteUser == "" || b.Env.PrometheusRemoteWritePassword == "") {
+		return fmt.Errorf("prometheus remote write username and password are required when remote write URL is set")
```
OliverTrautvetter: slightly confusing error that implies that both are not set (even if one of them is set)

maybe change it to something more general like:
"prometheus remote write username and password must both be set when remote write URL is specified"

internal/bootstrap/gcp/gcp_test.go
```diff
@@ -575,6 +575,47 @@ var _ = Describe("GCP Bootstrapper", func() {
...
+		Context("When Prometheus remote write URL is set but credentials are missing", func() {
+			BeforeEach(func() {
+				csEnv.PrometheusRemoteWriteURL = "https://prometheus.example.com/api/v1/write"
+			})
+			It("returns an error", func() {
+				err := bs.ValidateInput()
+				Expect(err).To(MatchError(ContainSubstring("prometheus remote write username and password are required when remote write URL is set")))
+			})
+		})
+		Context("When Prometheus remote write URL is set but only password is missing", func() {
```
OliverTrautvetter: counterpart for "URL set but only username is missing"

OliverTrautvetter (changes requested):
Some comments, looks good in general 👍

internal/installer/files/config_yaml.go
BBesrour: Unfortunately, the private installer code expects both to be under the `Password` field

OliverTrautvetter (approved):
looks good 👍

PR #455, opened 2026-05-27
feat(argocd): expose --force-conflicts, --values, and --repo flags with OCI support

schrodit:
- Add InstallChartOptions struct with ForceConflicts field to the HelmClient interface
- Replace hardcoded ForceConflicts=true in UpgradeChart with a configurable option
- Propagate ForceConflicts through install and upgrade paths in the ArgoCD installer
- Add --force-conflicts and --values/-f flags to the argocd CLI command
- Add --repo flag to configure the Helm chart repository (HTTP or OCI)
- Auto-detect OCI repos (oci:// prefix) and build the full chart reference accordingly

### DIFF
--- a/cli/cmd/argocd.go
+++ b/cli/cmd/argocd.go
@@ -23,6 +23,9 @@ type InstallArgoCDOpts struct {
 	GitPassword      string
 	RegistryPassword string
 	FullInstall      bool
+	ForceConflicts   bool
+	RepoURL          string
+	ValueFiles       []string
 }
 
 func (c *InstallArgoCDCmd) RunE(_ *cobra.Command, args []string) error {
@@ -39,7 +42,7 @@ func (c *InstallArgoCDCmd) RunE(_ *cobra.Command, args []string) error {
 			}
 		}
 	}
-	install, err := installer.NewArgoCD(c.Opts.Version, c.Opts.DatacenterId, c.Opts.RegistryPassword, c.Opts.GitPassword, c.Opts.FullInstall)
+	install, err := installer.NewArgoCD(c.Opts.Version, c.Opts.DatacenterId, c.Opts.RegistryPassword, c.Opts.GitPassword, c.Opts.FullInstall, c.Opts.ForceConflicts, c.Opts.RepoURL, c.Opts.ValueFiles)
 	if err != nil {
 		return fmt.Errorf("failed to initialize ArgoCD installer: %w", err)
 	}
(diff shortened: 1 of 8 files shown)


internal/installer/helm_client.go
```diff
@@ -122,7 +127,6 @@ func (h *helmClient) FindRelease(namespace, releaseName string) (*ReleaseInfo, e
 
 	listClient := action.NewList(env.actionConfig)
 	listClient.Filter = "^" + releaseName + "$"
-	listClient.Deployed = true
```
OliverTrautvetter, line 125: Why was `listClient.Deployed = true `removed? If it's a no-op (because NewList defaults to it), we should document that.

FindRelease only sees deployed releases. If a previous install ended in failed state, FindRelease returns nil and InstallChart will fail. So should we also set listClient.Failed = true?

internal/installer/argocd.go
```diff
@@ -62,18 +73,23 @@ func (a *ArgoCD) Install() error {
 
 	ctx := context.Background()
 
+	vals, err := (&values.Options{
+		ValueFiles: a.ValueFiles,
+		Values:     []string{"dex.enabled=false"},
+	}).MergeValues(getter.All(cli.New()))
```
OliverTrautvetter, line 76: Maybe an issue. `dex.enabled=false` cannot be overridden by any --values file. If it should be 'default that users can override' the order should swap.

And if it must be non-overridable we should add a comment

internal/installer/argocd.go
OliverTrautvetter: Pre existing dead code? settings and actionConfig are not used after that, right? I think we can remove that

cli/cmd/argocd.go
```diff
@@ -68,6 +71,9 @@ func AddArgoCDCmd(parentCmd *cobra.Command, opts *GlobalOptions) {
 	argocd.cmd.Flags().StringVar(&argocd.Opts.DatacenterId, "dc-id", "", "Codesphere Datacenter ID where this ArgoCD is installed")
 	argocd.cmd.Flags().StringVarP(&argocd.Opts.Version, "version", "v", "", "Version of the ArgoCD helm chart to install")
 	argocd.cmd.Flags().BoolVar(&argocd.Opts.FullInstall, "deploy-dc-config", false, "Install Codesphere-managed resources (AppProjects, Repo Creds, ...) after installing the chart")
+	argocd.cmd.Flags().StringArrayVarP(&argocd.Opts.ValueFiles, "values", "f", nil, "Specify values in a YAML file (can be specified multiple times)")
+	argocd.cmd.Flags().BoolVar(&argocd.Opts.ForceConflicts, "force-conflicts", false, "Force field ownership conflicts during upgrade (sets server-side apply ForceConflicts)")
+	argocd.cmd.Flags().StringVar(&argocd.Opts.RepoURL, "repo", "", "Helm chart repository URL; supports HTTP (default: https://argoproj.github.io/argo-helm) and OCI (e.g. oci://ghcr.io/argoproj/argo-helm)")
```
OliverTrautvetter, line 76: I think we should add a simple check in RunE validating that --repo starts with http://, https://, or oci:// when non-empty

internal/installer/argocd_test.go
```diff
@@ -186,4 +186,58 @@ var _ = Describe("ArgoCD.Install", func() {
 		})
 	})
 
+	Context("ForceConflicts", func() {
```
OliverTrautvetter, line 286: No tests for --values? We should check if the merge works

OliverTrautvetter (changes requested):
Some comments, looks good in general 👍

internal/installer/helm_client.go
schrodit: >FindRelease only sees deployed releases. If a previous install ended in failed state, FindRelease returns nil and InstallChart will fail. So should we also set listClient.Failed = true

yes lets do that And I will add a comment why we need it

internal/installer/argocd.go
schrodit: good catch. it should be overridable.

internal/installer/argocd.go
schrodit: ah yes, its actually unused. though go would catch that. but anyway removed it

OliverTrautvetter (approved):
lgtm 🚀

PR #457, opened 2026-05-27
feat(installer): support templated install configs from vault

schrodit:
## Summary

- Add Go template rendering for install configs with dynamic `{{ secret "..." }}` lookups backed by `prod.vault.yaml`
  - I decided to not use a simple struct like `{{ vault.mySecret.password }}` because using a function make it more generic to support other stores like vault ehre secrets are rather fetched async.
- Add a generic `configtemplating` package and a vault-backed installer secret store
  - The current vault is in a separate package to be able to switch the store with a different one e.g. openbao.
- Support explicit vault path handling via `oms install codesphere --vault`
- Render install configs to temporary `0600` files and warn when `secrets.baseDir` differs from the passed vault directory

## Example

Use the newly added `oms config template` command to test it out locally.

```yaml
# Example config.yaml fragment using secrets from prod.vault.yaml.
#
# Rendered by:
#   oms install codesphere --config config.yaml --vault prod.vault.yaml --priv-key age_key.txt ...
#
# Secret values are read dynamically from prod.vault.yaml and are not stored in
# plaintext in config.yaml.

secrets:
  baseDir: ./secrets

codesphere:
  override:
    global:
      license:
        key: '{{ secret "codesphereLicenseKey" }}'

postgres:
  override:
    auth:
      username: '{{ secret "postgresAdmin" "fields.username" }}'
      password: '{{ secret "postgresAdmin" "fields.password" }}'
```


cli/cmd/template_config.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return nil
+}
+
+func AddConfigTemplateCmd(parentCmd *cobra.Command, opts *GlobalOptions) {
+	templateCmd := &ConfigTemplateCmd{
+		cmd: &cobra.Command{
+			Use:   "template",
+			Short: "Render a config.yaml template using secrets from a vault file",
```
joka134, line 50: Should we add a warning here as it will print secrets?

cli/cmd/config.go
joka134, line 1: There's also `oms init install-config`. We should move that here as well to `oms config init`. We can create a ticket when this is merged

joka134: Plus having `oms config ...` is kind of breaking with the current style, which would be more like `oms template config`. 

See current commands:
```
Available Commands:
  beta        Commands for early testing
  build       Build and push images to a registry
  completion  Generate the autocompletion script for the specified shell
  create      Create resources for Codesphere
  download    Download resources available through OMS
  help        Help about any command
  init        Initialize configuration files
  install     Install Codesphere and other components
  licenses    Print license information
  list        List resources available through OMS
  register    Register a new API key
  revoke      Revoke resources available through OMS
  smoketest   Run smoke tests for Codesphere components
  update      Update OMS related resources
  version     Print version
```

I personally like `oms config ...` more, but we should be consistent

internal/installer/vault_templating_secret_store.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return NewVaultTemplatingSecretStore(vault), nil
+}
+
+func (s *VaultTemplatingSecretStore) LookupSecret(name string, selector ...string) (string, error) {
+	if s == nil {
+		return "", errors.New("vault secret store is not configured")
+	}
```
joka134: Why are we checking this explicitly here? We haven't done this in any of the other methods. Plus if we can call this method, `s` should not be nil

internal/configtemplating/config_template.go
```diff
@@ -0,0 +1,81 @@
...
+	"os"
+	"text/template"
+)
+
+// SecretStore is the template-facing abstraction for secret backends.
+type SecretStore interface {
+	LookupSecret(name string, selector ...string) (string, error)
+}
+
+func RenderInstallConfigTemplate(data []byte, store SecretStore) ([]byte, error) {
```
joka134, line 21: Let's add some short comments on all exported functions, so that no one needs to jump into the implementation to see what is returned here.

internal/installer/vault_templating_secret_store.go
```diff
@@ -0,0 +1,156 @@
...
+	"go.yaml.in/yaml/v3"
+)
+
+type VaultTemplatingSecretStore struct {
+	vault      *files.InstallVault
+	vaultPath  string
+	ageKeyPath string
+}
+
+func NewVaultTemplatingSecretStore(vault *files.InstallVault) *VaultTemplatingSecretStore {
```
joka134, line 25: Comments on exported functions and structs are missing

internal/installer/vault_templating_secret_store.go
```diff
@@ -0,0 +1,156 @@
...
+	return NewVaultTemplatingSecretStore(vault), nil
+}
+
+func (s *VaultTemplatingSecretStore) LookupSecret(name string, selector ...string) (string, error) {
+	if s == nil {
+		return "", errors.New("vault secret store is not configured")
+	}
+	if s.vault == nil {
+		if s.vaultPath == "" {
+			return "", errors.New("vault secret store is not configured")
```
joka134: Same error as above? Would specify more like: `error initializing vault: vaultPath not set`.
To get that prefix on the following error as well it might even make sense to split this "init" part into a new function.

internal/installer/vault_templating_secret_store.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		vault, err := LoadVaultData(s.vaultPath, s.ageKeyPath)
+		if err != nil {
+			return "", err
```
joka134: no specific error?

internal/installer/vault_templating_secret_store.go
```diff
@@ -0,0 +1,156 @@
...
+	if len(doc.Content) == 0 {
+		return false, nil
+	}
+
+	root := doc.Content[0]
+	if root.Kind != yaml.MappingNode {
+		return false, nil
+	}
+
+	for i := 0; i+1 < len(root.Content); i += 2 {
```
joka134, line 163: `i += 2` because this is actually iterating the lines or how is this working? Small comment about this logic would be great

internal/installer/config_template_test.go
```diff
@@ -0,0 +1,125 @@
...
+		vaultPath := filepath.Join(tempDir, "prod.vault.yaml")
+		plaintextVaultPath := filepath.Join(tempDir, "prod.vault.plain.yaml")
+		ageKeyPath := filepath.Join(tempDir, "age_key.txt")
+
+		Expect(os.WriteFile(configPath, []byte(`secrets:
+  baseDir: "`+tempDir+`"
+codesphere:
+  override:
+    apiToken: "{{ secret "apiToken" }}"
+`), 0644)).To(Succeed())
```
joka134: weird code formatting. can you move out the payload into a new variable and use something like `fmt.Sprintf` with a multi-line string?

internal/configtemplating/config_template.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, fmt.Errorf("failed to render config template: %w", err)
+	}
+
+	return rendered.Bytes(), nil
+}
+
+// RenderConfigFileToTempIfNeeded renders configPath with store into a temporary
+// 0600 YAML file and returns that path plus a cleanup function. Callers should
+// pass the returned path to downstream code and defer cleanup immediately.
+func RenderConfigFileToTempIfNeeded(configPath string, store SecretStore) (string, func(), error) {
```
joka134: What do you mean with `IfNeeded`?  isn't this always going to create a tmp folder?

cli/cmd/install_codesphere.go
```diff
@@ -98,10 +102,27 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
...
 
+	originalConfig := c.Opts.Config
+	cleanup := func() {}
+	if c.Opts.Vault != "" {
+		store := installer.NewLazyVaultTemplatingSecretStore(c.Opts.Vault, c.Opts.PrivKey)
+		renderedConfig, renderCleanup, err := configtemplating.RenderConfigFileToTempIfNeeded(c.Opts.Config, store)
+		if err != nil {
+			return fmt.Errorf("failed to render config template: %w", err)
+		}
+		cleanup = renderCleanup
```
joka134, line 113: Why don't you call cleanup here directly?

joka134 (changes requested):
nice feature! 🚀  Left some comments. Let's discuss it if you want

internal/installer/vault_templating_secret_store.go
schrodit: yes make sense. I thought that I added exactly that already. let me check

cli/cmd/install_codesphere.go
schrodit: you mean the defer?
Because the config is used in the installer so it must only be deleted once the installer is done. And I think we call the defer outside of the if because otherwise defer would be called at the end of the if

cli/cmd/config.go
schrodit: I now rather used the style that we currently have. With this I would not move the config init here

cli/cmd/install_codesphere.go
joka134: Defer is called when the function returns, so that should be fine.

> A defer statement pushes a function call onto a list. The list of saved calls is executed after the surrounding function returns.
https://go.dev/blog/defer-panic-and-recover

internal/installer/vault_templating_secret_store.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return NewVaultTemplatingSecretStore(vault), nil
+}
+
+// LookupSecret returns the value of the named secret, optionally narrowed by a
+// field selector (e.g. "password", "file.content"). The vault is loaded lazily
+// on first use when the store was created without a preloaded vault.
+func (s *VaultTemplatingSecretStore) LookupSecret(name string, selector ...string) (string, error) {
+	if err := s.ensureVault(); err != nil {
+		return "", err
```
joka134:
```suggestion
		return "", fmt.Errorf("error ensuring the vault: %w", err)
```
then you can remove the duplicated prefix from the errors inside the `ensureVault` function

internal/installer/config_template_test.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		tempDir := GinkgoT().TempDir()
+		configPath := filepath.Join(tempDir, "config.yaml")
+		vaultPath := filepath.Join(tempDir, "prod.vault.yaml")
+		plaintextVaultPath := filepath.Join(tempDir, "prod.vault.plain.yaml")
+		ageKeyPath := filepath.Join(tempDir, "age_key.txt")
+
+		configYaml := fmt.Sprintf(`secrets:
+  baseDir: %q
```
joka134, line 87: indendation has to be like this, right? Else yaml adds more spaces than needed? 😄

cli/cmd/template_config.go
```diff
@@ -0,0 +1,91 @@
...
+
+func AddTemplateConfigCmd(parentCmd *cobra.Command, opts *GlobalOptions) {
+	configCmd := &TemplateConfigCmd{
+		cmd: &cobra.Command{
+			Use:   "config",
+			Short: "Render a config.yaml template using secrets from a vault file",
+			Long: io.Long(`Render a config.yaml template using secrets from a prod.vault.yaml file.
+
+This command prints the rendered configuration to stdout so templating can be tested without running an installation.`),
+			Example: formatExamples("template config", []io.Example{
```
joka134, line 68: Can we add some examples how the syntax in the config.yaml needs to look like?

joka134 (changes requested):
Nice! added some optional comments. But adding some examples to the CLI would simplify the usage.

internal/installer/config_template_test.go
schrodit: moved the `secrets:` to the next line to make the file easier to read. but otherwise the format should be correct?

NOTICE
```diff
@@ -67,7 +67,7 @@ License URL: https://gitea.com/gitea/go-sdk/src/tag/gitea/v0.25.1/gitea/LICENSE
 Module: dario.cat/mergo
 Version: v1.0.2
 License: BSD-3-Clause
-License URL: https://github.com/imdario/mergo/blob/v1.0.2/LICENSE
+License URL: Unknown
```
joka134, line 70: Oh no, not again

joka134 (approved):
🚀 Nice!

NOTICE
joka134: Why is the license updated here, the version hasn't changed 🤷

schrodit: but only that one module. locally is has the same issue. so something we have to check in all of oms

schrodit: will do in a follow up pr

PR #461, opened 2026-06-05
fix: support unenrypted vault files for templating

joka134:
This is hotfix to support recovery of bootstrapped GCP projects, where the vault has been encrypted before loading the vault for templating. We should discuss if we want to change the way of recovering the vault (decrypt on jumphost and download vs. download age_key.txt and decrypt when needed).


NJona (approved):
lgtm as a hotfix until we decide on how we want to handle it.

PR #463, opened 2026-06-09
feat(config.yaml): add operations.skip option to config.yaml

bachgg:
One can now add steps to skip into `config.yaml` under `operations.skip` as a list of string. The steps to be skipped can be defined both in `config.yaml` and cli args, and all those steps will be merged. An interactive prompt can be turned on by running the cli with `--auto-approve=false` to confirm the steps to be skipped.

### DIFF
--- a/cli/cmd/install_codesphere.go
+++ b/cli/cmd/install_codesphere.go
@@ -239,10 +239,16 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
 	archivePath := filepath.Join(pm.GetWorkDir(), "deps.tar.gz")
 
 	cmdArgs := []string{installerPath, "--archive", archivePath, "--config", c.Opts.Config, "--privKey", c.Opts.PrivKey}
-	if len(c.Opts.SkipSteps) > 0 {
-		for _, step := range c.Opts.SkipSteps {
-			cmdArgs = append(cmdArgs, "--skipStep", step)
-		}
+	if len(config.Operations.Skip) > 0 && len(c.Opts.SkipSteps) > 0 {
+		return fmt.Errorf("skipped steps can either be defined in config.yaml under `.operations.skip` or supplied via `--skip-step` flag, but not both")
+	}
+
+	for _, step := range config.Operations.Skip {
+		cmdArgs = append(cmdArgs, "--skipStep", step)
+	}
+
+	for _, step := range c.Opts.SkipSteps {
+		cmdArgs = append(cmdArgs, "--skipStep", step)
 	}
 
 	if c.Opts.CodesphereOnly {
--- a/cli/cmd/install_codesphere.go
+++ b/cli/cmd/install_codesphere.go
@@ -239,12 +239,14 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
 	archivePath := filepath.Join(pm.GetWorkDir(), "deps.tar.gz")
 
 	cmdArgs := []string{installerPath, "--archive", archivePath, "--config", c.Opts.Config, "--privKey", c.Opts.PrivKey}
-	if len(config.Operations.Skip) > 0 && len(c.Opts.SkipSteps) > 0 {
-		return fmt.Errorf("skipped steps can either be defined in config.yaml under `.operations.skip` or supplied via `--skip-step` flag, but not both")
-	}
+	if config.Operations != nil {
+		if len(config.Operations.Skip) > 0 && len(c.Opts.SkipSteps) > 0 {
+			return fmt.Errorf("skipped steps can either be defined in config.yaml under `.operations.skip` or supplied via `--skip-step` flag, but not both")
+		}
 
-	for _, step := range config.Operations.Skip {
-		cmdArgs = append(cmdArgs, "--skipStep", step)
+		for _, step := range config.Operations.Skip {
+			cmdArgs = append(cmdArgs, "--skipStep", step)
+		}
 	}
 
 	for _, step := range c.Opts.SkipSteps {
--- a/cli/cmd/install_codesphere.go
+++ b/cli/cmd/install_codesphere.go
@@ -267,24 +267,21 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
 		executableSteps[step] = false
 	}
 
+	executedSteps := []string{}
 	for step, executed := range executableSteps {
 		if !executed {
 			cmdArgs = append(cmdArgs, "--skipStep", step)
-		}
-	}
-
-	executedSteps := []string{}
-	for step, executed := range executableSteps {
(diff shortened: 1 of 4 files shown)


cli/cmd/install_codesphere.go
```diff
@@ -239,10 +239,16 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
 	archivePath := filepath.Join(pm.GetWorkDir(), "deps.tar.gz")
 
 	cmdArgs := []string{installerPath, "--archive", archivePath, "--config", c.Opts.Config, "--privKey", c.Opts.PrivKey}
-	if len(c.Opts.SkipSteps) > 0 {
-		for _, step := range c.Opts.SkipSteps {
-			cmdArgs = append(cmdArgs, "--skipStep", step)
-		}
+	if len(config.Operations.Skip) > 0 && len(c.Opts.SkipSteps) > 0 {
```
NautiluX: couldn't Operations be `nil`?

bachgg: Added a nil check

cli/cmd/install_codesphere.go
```diff
@@ -239,10 +239,16 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
...
+		return fmt.Errorf("skipped steps can either be defined in config.yaml under `.operations.skip` or supplied via `--skip-step` flag, but not both")
```
NautiluX: why not both? We could build a union of both I guess?

internal/installer/files/config_yaml.go
```diff
@@ -62,6 +62,11 @@ type RootConfig struct {
 	MetalLB                *MetalLBConfig                `yaml:"metallb,omitempty"`
 	Codesphere             CodesphereConfig              `yaml:"codesphere"`
 	ManagedServiceBackends *ManagedServiceBackendsConfig `yaml:"managedServiceBackends,omitempty"`
+	Operations             *OperationsConfig             `yaml:"operations,omitempty"`
```
NautiluX, line 65: did you verify the PC installer ignores this setting?
wondering if long-term `operations` with a list of steps being`skip`ped is the right format. How about something like
```
components:
    exclude:
    - sops
    - docker
```
which would allow us to use `include` in the future?

cli/cmd/install_codesphere.go
bachgg: I think it's better to only define them in `config.yaml` so that each run config is persisted with git.

cli/cmd/install_codesphere.go
```diff
@@ -239,12 +239,56 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
...
+	executedSteps := []string{}
+	for component, _ := range skippableSteps {
+		executedSteps = append(executedSteps, component)
+	}
+
+	prompt := installer.NewPrompter(true)
+	msg := fmt.Sprintf("The following steps will be executed: %s. Type \"yes\" to continue.", strings.Join(executedSteps, ", "))
+	if prompt.String(msg, "no") != "yes" {
+		return fmt.Errorf("Installation aborted")
+	}
```
NautiluX: I think we need an option to do this noninteractively

cli/cmd/install_codesphere.go
```diff
@@ -239,12 +239,56 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
...
+		"load-container-images": {},
+		"sops":                  {},
+		"docker":                {},
+		"postgres":              {},
+		"ceph":                  {},
+		"kubernetes":            {},
+		"set-up-cluster":        {},
+		"codesphere":            {},
+		"ms-backends":           {},
+	}
```
NautiluX: why not `executedSteps` as `map[string]bool`? That way you can set skipped steps to false

cli/cmd/install_codesphere.go
```diff
@@ -239,12 +239,56 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
...
+
+		for _, step := range config.Operations.Skip {
+			skippedSteps = append(skippedSteps, step)
+		}
+	}
+
 	if len(c.Opts.SkipSteps) > 0 {
 		for _, step := range c.Opts.SkipSteps {
-			cmdArgs = append(cmdArgs, "--skipStep", step)
+			skippedSteps = append(skippedSteps, step)
```
NautiluX: guess we could update the map here directly no?

cli/cmd/install_codesphere.go
```diff
@@ -239,12 +239,56 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
...
+	}
+	skippedSteps := []string{}
+
+	if config.Operations != nil {
+		if len(config.Operations.Skip) > 0 && len(c.Opts.SkipSteps) > 0 {
+			return fmt.Errorf("skipped steps can either be defined in config.yaml under `.operations.skip` or supplied via `--skip-step` flag, but not both")
+		}
+
+		for _, step := range config.Operations.Skip {
+			skippedSteps = append(skippedSteps, step)
```
NautiluX: guess we could update the map here directly no?

cli/cmd/install_codesphere.go
```diff
@@ -239,12 +239,56 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
...
+		"sops":                  {},
+		"docker":                {},
+		"postgres":              {},
+		"ceph":                  {},
+		"kubernetes":            {},
+		"set-up-cluster":        {},
+		"codesphere":            {},
+		"ms-backends":           {},
+	}
+	skippedSteps := []string{}
```
NautiluX: Do we need both the map and the slice of skippedSteps?

cli/cmd/install_codesphere.go
```diff
@@ -239,12 +242,48 @@ func (c *InstallCodesphereCmd) ExtractAndInstall(pm installer.PackageManager, cm
...
+	executedSteps := []string{}
+	for step, executed := range executableSteps {
+		if !executed {
 			cmdArgs = append(cmdArgs, "--skipStep", step)
+		} else {
+			executedSteps = append(executedSteps, step)
 		}
 	}
 
+	sort.Strings(executedSteps)
```
NautiluX, line 279: I guess the order of the map is the order in which the steps are executed, wouldn't it be more helpful to the user to keep this order? I know the order is not in the control of OMS, but at least a rough order of execution would be nice I think.

NautiluX (approved):

cli/cmd/install_codesphere.go
bachgg: Map keys are unordered, so it's not guaranteed you get the same output every time if it's not sorted

PR #473, opened 2026-06-11
feat: force-conflicts in pc-application install

liamhess:
We had mutliple failed master deploys with the pc-application install conflicting on SSA managed fields.

Since the oms installer should have higher authority than the argo app controller it should be able to force field ownership to itself in the install step.

So we expose a force-conflicts option that alreay exists in the helm client to be used in the oms cli.

### DIFF
--- a/cli/cmd/pc_apps.go
+++ b/cli/cmd/pc_apps.go
@@ -22,9 +22,10 @@ type InstallPCAppsCmd struct {
 
 type InstallPCAppsOpts struct {
 	*GlobalOptions
-	Version     string
-	Namespace   string
-	ValuesFiles []string
+	Version        string
+	Namespace      string
+	ValuesFiles    []string
+	ForceConflicts bool
 }
 
 func (c *InstallPCAppsCmd) RunE(cmd *cobra.Command, args []string) error {
@@ -43,6 +44,7 @@ func (c *InstallPCAppsCmd) RunE(cmd *cobra.Command, args []string) error {
 		c.Opts.Version,
 		c.Opts.Namespace,
 		c.Opts.ValuesFiles,
+		c.Opts.ForceConflicts,
 	)
 	if err != nil {
 		return fmt.Errorf("failed to initialize pc-apps installer: %w", err)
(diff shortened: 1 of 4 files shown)


internal/installer/pc_apps_test.go
```diff
@@ -215,11 +215,32 @@ var _ = Describe("PCApps.Install", func() {
...
-			pcApps = installer.NewPCAppsForTesting(helmMock, fakeClient, version, namespace, []string{badFile})
+			pcApps = installer.NewPCAppsForTesting(helmMock, fakeClient, version, namespace, []string{badFile}, false)
 
 			err := pcApps.Install(context.Background())
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(ContainSubstring("loading values files"))
 		})
 	})
+
+	Context("ForceConflicts", func() {
```
OliverTrautvetter, line 235: The ForceConflicts test only covers the fresh install path -> fake client has no existing Helm release -> UpgradeChart with InstallIfNotExist: true falls through to InstallChart

 So please add a second It block that seeds a fake release

internal/installer/pc_apps_test.go
OliverTrautvetter, line 89: It uses mock.Anything for all UpgradeChart parameters. 

If ForceConflicts changes from false to true the test would still pass silently

cli/cmd/pc_apps.go
OliverTrautvetter, line 76: we should add some example for the new flag

OliverTrautvetter (changes requested):
Looks very good in general, some small comments 👍

internal/installer/pc_apps_test.go
liamhess: True, agree that this would be nice, however the distinction whether we upgrade or install happens inside of `helmClient.UpgradeChart`
And what we test / observer here is that `helmClient.UpgradeChart` gets the `ForceConflicts: true`
So I think the test would rather belong into `helm_client_test.go` and not here.
If you want I can add that file (`helm_client_test.go`) and add a test there that checks this.
But I don't think that this test belongs here, what do you think?
Can very well be that I am missing something as I'm not super proficient with go😅

internal/installer/pc_apps_test.go
liamhess: will expect it to be false

cli/cmd/pc_apps.go
liamhess: added

OliverTrautvetter (approved):
🚀

internal/installer/pc_apps_test.go
liamhess: Agreed on Slack that we don't add extra test/file for now

PR #474, opened 2026-06-11
fix(installer/node): fs.inotify.max_user_instances not configured

gnarlex:
Looked at the node, and the `inotify.max_user_watches` value is correctly set, but the `inotify.max_user_instances`  param is still set to the (default) 128  instead of `8192` we require.

Set `inotify.max_user_instances` to `8192` during node provisioning.

Task-ID: CU-869d13eeg

### DIFF
--- a/internal/installer/node/node.go
+++ b/internal/installer/node/node.go
@@ -361,7 +361,7 @@ func (n *Node) isSysctlActive(key, expected string) bool {
 }
 
 // configureSysctlLine appends a specific line to /etc/sysctl.conf and applies the settings on the remote node via SSH
-func (n *Node) configureSysctlLines(line string) error {
+func (n *Node) configureSysctlLine(line string) error {
 	if !n.hasSysctlLine(line) {
 		cmd := fmt.Sprintf("echo '%s' | sudo tee -a /etc/sysctl.conf", line)
 		if err := n.RunSSHCommand("root", cmd); err != nil {
--- a/internal/installer/node/node.go
+++ b/internal/installer/node/node.go
@@ -274,11 +274,23 @@ func (n *Node) EnableRootLogin() error {
 }
 
 func (n *Node) HasInotifyWatchesConfigured() bool {
-	return n.hasSysctlLine("fs.inotify.max_user_watches=1048576") && n.isSysctlActive("fs.inotify.max_user_watches", "1048576")
+	return n.hasSysctlLine("fs.inotify.max_user_watches=1048576") &&
+	    n.isSysctlActive("fs.inotify.max_user_watches", "1048576") &&
+	    n.hasSysctlLine("fs.inotify.max_user_instances=8192") &&
+	    n.isSysctlActive("fs.inotify.max_user_instances", "8192")
 }
 
 func (n *Node) ConfigureInotifyWatches() error {
-	return n.configureSysctlLine("fs.inotify.max_user_watches=1048576")
+	lines := []string{
+		"fs.inotify.max_user_instances=8192",
+		"fs.inotify.max_user_watches=1048576",
+	}
+	for _, ln := range lines {
+        err := n.configureSysctlLine(ln)
+		if err != nil {
+			return fmt.Errorf("failed to configure systctl line '%s': %w", ln, err)
+		}
+	}
 }
 
 func (n *Node) HasMemoryMapConfigured() bool {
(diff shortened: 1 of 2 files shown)


internal/installer/node/node.go
```diff
@@ -274,11 +274,23 @@ func (n *Node) EnableRootLogin() error {
...
+	lines := []string{
+		"fs.inotify.max_user_instances=8192",
+		"fs.inotify.max_user_watches=1048576",
+	}
+	for _, ln := range lines {
+        err := n.configureSysctlLine(ln)
+		if err != nil {
+			return fmt.Errorf("failed to configure systctl line '%s': %w", ln, err)
+		}
+	}
```
gnarlex:
```suggestion
	}
	return nil
```

internal/installer/node/node.go
```diff
@@ -274,11 +274,24 @@ func (n *Node) EnableRootLogin() error {
...
 }
 
 func (n *Node) ConfigureInotifyWatches() error {
-	return n.configureSysctlLine("fs.inotify.max_user_watches=1048576")
+	lines := []string{
+		"fs.inotify.max_user_instances=8192",
+		"fs.inotify.max_user_watches=1048576",
+	}
+	for _, ln := range lines {
+        err := n.configureSysctlLine(ln)
```
OliverTrautvetter: `configureSysctlLine` always calls `sudo sysctl -p`, so also if line was already present. It is also done twice now, 1 with _instances and 1 with _watche.

So can we do that only at the end?

internal/installer/node/node.go
```diff
@@ -274,11 +274,24 @@ func (n *Node) EnableRootLogin() error {
...
+		if err != nil {
+			return fmt.Errorf("failed to configure systctl line '%s': %w", ln, err)
```
OliverTrautvetter:
```suggestion
			return fmt.Errorf("failed to configure sysctl line '%s': %w", ln, err)
```

typo

internal/installer/node/node.go
```diff
@@ -274,11 +274,24 @@ func (n *Node) EnableRootLogin() error {
 }
 
 func (n *Node) HasInotifyWatchesConfigured() bool {
-	return n.hasSysctlLine("fs.inotify.max_user_watches=1048576") && n.isSysctlActive("fs.inotify.max_user_watches", "1048576")
+	return n.hasSysctlLine("fs.inotify.max_user_watches=1048576") &&
+	    n.isSysctlActive("fs.inotify.max_user_watches", "1048576") &&
+	    n.hasSysctlLine("fs.inotify.max_user_instances=8192") &&
+	    n.isSysctlActive("fs.inotify.max_user_instances", "8192")
```
OliverTrautvetter: line 277, 278, 279 and 289, if I save it auto formats them, I think there are spaces instead of 2 tabs to put them in place

internal/installer/node/node.go
```diff
@@ -274,11 +274,24 @@ func (n *Node) EnableRootLogin() error {
...
 }
 
 func (n *Node) ConfigureInotifyWatches() error {
-	return n.configureSysctlLine("fs.inotify.max_user_watches=1048576")
+	lines := []string{
+		"fs.inotify.max_user_instances=8192",
+		"fs.inotify.max_user_watches=1048576",
+	}
```
OliverTrautvetter, line 287:
```suggestion
	lines := []string{
		"fs.inotify.max_user_watches=1048576",
		"fs.inotify.max_user_instances=8192",
	}
```
this was inconsistent, HasInotifyWatchesConfigured checks _watches before _instances

OliverTrautvetter (changes requested):
Looks very good in general, some small comments 👍

internal/installer/node/node.go
gnarlex: Why is CI green (lint)? Does it not check if gofmt changes?

OliverTrautvetter: Not sure, looks like lint doesn't care if it is 4 spaces or 1 tab

internal/installer/node/node.go
gnarlex: We're already calling sysctl multiple times. Once for inotify, once for vm memory map config.

Maybe it makes sense to just maintain a list of sysctl lines, and write them into the file. IMHO the sysctl part of the code seems a bit overengineered, trying to double- and triple check every sysctl config, specific to what it's trying to configure.

gnarlex: Will change `configureSysctlLine` to accept a list of lines, so that it only runs `sysctl -p` once at the end, but as I mentioned above: We may want to simplify the sysctl reconciliation in general.

OliverTrautvetter (approved):
Very nice 🚀

PR #479, opened 2026-06-12
feat(installer): add secrets package with idempotent secret generation

schrodit:
Introduce internal/installer/secrets as the single source of truth for all private-cloud secret generation:

- EnsureAuthKeys: RSA-4096 token key pair + EC P-256 domain-auth key pair (PKCS8/SPKI PEM), each checked independently so a partial vault can be completed without regenerating existing keys
- EnsureServiceAccountTokens: RS512 JWTs (1-year expiry) for all 10 Codesphere internal service accounts, signed with tokenPrivateKey
- EnsureIngressCA / EnsureCephSSHKeys: self-signed CA and SSH key pair
- EnsurePostgresSecrets: CA, server certs, passwords, and per-service passwords including the previously missing usageAggregationRefresher and usageAggregationReader credentials
- EnsureMounterHmacSecret: migrates legacy hmac-secret or creates new
- EnsureDefaultSecrets: dummy values for optional Helm chart secrets

EnsureSecrets orchestrates all of the above; GenerateSecrets delegates to it entirely and then calls MergeVaultIntoConfig to sync vault entries back into config struct fields.

Also fix a latent nil-pointer in MergeVaultIntoConfig when writing GitHub OAuth secrets into an unconfigured GitProviders struct.

The `oms install codesphere` command is currently not touched. There will be a dedicated `oms update secrets` command in a follow up PR to integrate it into the current prod install workflow


internal/installer/secrets/secrets.go
```diff
@@ -0,0 +1,423 @@
...
+const serviceAccountTokenExpiry = 365 * 24 * time.Hour
+
+// codesphereServiceUsers lists all internal service accounts and the vault key for their token.
+var codesphereServiceUsers = []serviceUser{
+	{tokenName: "authServiceUserToken", serviceID: "auth-service", email: "auth.service@codesphere.com"},
+	{tokenName: "paymentServiceUserToken", serviceID: "payment-service", email: "payment.service@codesphere.com"},
+	{tokenName: "publicApiServiceUserToken", serviceID: "public-api-service", email: "public.api.service@codesphere.com"},
+	{tokenName: "deploymentServiceUserToken", serviceID: "deployment-service", email: "deployment.service@codesphere.com"},
+	{tokenName: "marketplaceServiceUserToken", serviceID: "marketplace-service", email: "marketplace.service@codesphere.com"},
+	{tokenName: "errorPageServerUserToken", serviceID: "error-page-server", email: "error.page.servere@codesphere.com"},
```
OliverTrautvetter: typo "servere" in email

internal/installer/secrets/crypto.go
```diff
@@ -0,0 +1,157 @@
...
+	return string(privBytes), string(ssh.MarshalAuthorizedKey(sshPub)), nil
+}
+
+func generatePassword(length int) string {
+	b := make([]byte, length)
+	if _, err := rand.Read(b); err != nil {
+		return base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%d", time.Now().UnixNano())))[:length]
+	}
+	return base64.StdEncoding.EncodeToString(b)[:length]
+}
```
OliverTrautvetter: If rand.Read is failing the system is in a very broken state. I am not sure if we should return a deterministic value like the predictable time as password. In any case the function should return the error instead of swallowing it

internal/installer/secrets/crypto.go
```diff
@@ -0,0 +1,157 @@
...
+
+func generateSSHKeyPair() (privateKey, publicKey string, err error) {
+	key, err := rsa.GenerateKey(rand.Reader, 4096)
+	if err != nil {
+		return "", "", err
+	}
+
+	privBytes := pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(key),
```
OliverTrautvetter: All other use PKCS8 not PKCS1. I think this needs an update

OliverTrautvetter: I still find other code outside the changes that still uses PKCS1, update that too?

internal/installer/config_manager.go
OliverTrautvetter: We should add tokenPrivateKey and tokenPublicKey

internal/installer/secrets/secrets.go
```diff
@@ -0,0 +1,423 @@
...
+
+// EnsureServiceAccountTokens signs RS512 JWTs for all Codesphere internal service accounts
+// and stores them in vault. Requires tokenPrivateKey to already be present (call EnsureAuthKeys
+// first). Idempotent: skips if authServiceUserToken already exists.
+func EnsureServiceAccountTokens(vault *files.InstallVault) error {
+	if vault.GetSecret("authServiceUserToken") != nil {
+		return nil
+	}
+
+	privKeyEntry := vault.GetSecret("tokenPrivateKey")
```
OliverTrautvetter, line 85: What if authServiceUserToken and or tokenPrivateKey are chnaged? This would prevent regeneration

internal/installer/secrets/crypto.go
```diff
@@ -0,0 +1,157 @@
+// Copyright (c) Codesphere Inc.
```
OliverTrautvetter, line 1: secrets/crypto.go duplicates generateCA, generateServerCertificate, generateSSHKeyPair, generatePassword, and validateCertKeyPair from installer/crypto.

This is hard for maintenance. bug fixes / updates to one won't propagate to the other.

So either:
- the old public functions delegates to the new package
- just update update_install_config.go to use the new secrets package directly

OliverTrautvetter (changes requested):

internal/installer/secrets/crypto.go
schrodit: actually those create real passwords, so we should rather create secure ones.
If the password is irrelevant then we anyway only return a `dummy` password.

>In any case the function should return the error instead of swallowing it

yes make sense

internal/installer/secrets/crypto.go
schrodit: don't get your comment.
Here we rather parse the privte key and this function bascially supports both versions.
So why do we need to update the others?

OliverTrautvetter: My concern is parseCAKeyAndCert in installer/crypto.go (which uses ParsePKCS1PrivateKey) being unable to parse PKCS8 keys generated by the new generateCA/generateServerCertificate. If update_install_config.go tries to regenerate certs from a vault populated by this new code, I think it will break.

internal/installer/secrets/crypto.go
schrodit: move everything to secrets now.
Seems more fitting

internal/installer/secrets/crypto.go
schrodit: done returned an error

internal/installer/secrets/crypto.go
schrodit: as said in the other package.
Now both the use from `installer/secrets` which supports both so shouldn't be an issue anymore

internal/installer/config_manager.go
OliverTrautvetter: not added yet

internal/installer/secrets/secrets.go
OliverTrautvetter: We now have the new GenerateECDSAKeyPair in crypto.go and the new generateECP256PKCS8KeyPair in secrets.go.

func generateECP256PKCS8KeyPair() (privatePEM, publicPEM string, err error) {

func GenerateECDSAKeyPair() (privateKey string, publicKey string, err error) {

So 2 EC key generators in the same package produce different PEM formats

OliverTrautvetter (changes requested):
Looks much better now. 1 more comment and tokenPrivateKey and tokenPublicKey are not added to requiredSecrets. (tests are broken now)

internal/installer/secrets/secrets.go
schrodit: now done, only one is left

OliverTrautvetter (approved):
Looks good now 🚀

internal/installer/config_manager.go
```diff
@@ -175,7 +175,7 @@ func (g *InstallConfig) ValidateVault() []string {
 	}
 
 	errors := []string{}
-	requiredSecrets := []string{"cephSshPrivateKey", "selfSignedCaKeyPem", "domainAuthPrivateKey", "domainAuthPublicKey"}
+	requiredSecrets := []string{"tokenPrivateKey", "tokenPublicKey", "cephSshPrivateKey", "selfSignedCaKeyPem", "domainAuthPrivateKey", "domainAuthPublicKey"}
```
NautiluX, line 178: can we reference them using the constants here as well?

internal/installer/config_manager_secrets.go
```diff
@@ -4,89 +4,18 @@
...
-	services := []string{"auth", "deployment", "ide", "marketplace", "payment", "public_api", "team", "workspace"}
-	config.Postgres.UserPasswords = make(map[string]string)
-	for _, service := range services {
-		config.Postgres.UserPasswords[service] = GeneratePassword(32)
-	}
-
-	return nil
+	// Sync vault → config struct fields so that ExtractVault and any code reading
+	// yaml:"-" fields (private keys, passwords) sees the generated values.
+	return g.MergeVaultIntoConfig()
```
NautiluX, line 20: do we still need this if we always reference the vault object directly?

internal/installer/files/config_yaml.go
```diff
@@ -968,7 +978,7 @@ func (c *RootConfig) addPostgresSecrets(vault *InstallVault) {
 		},
 	})
 
-	services := []string{"auth", "deployment", "ide", "marketplace", "payment", "public_api", "team", "workspace"}
+	services := []string{"auth", "deployment", "ide", "marketplace", "payment", "public_api", "team", "workspace", "usageAggregationRefresher", "usageAggregationReader"}
```
NautiluX, line 981: is this the same struct as in config_manager.go? Do we need both?

NautiluX (commented):
I like this approach, do we still need the merged secrets inside of the config struct now?

PR #483, opened 2026-06-13
feat(local): Add argocd deployment to local bootstrap

schrodit:
Adds the argocd deployment to the local installer.
Refactors some parts to be better reusable within the project e.g. bom or argocd package


internal/bootstrap/local/local.go
```diff
@@ -214,6 +225,22 @@ func (b *LocalBootstrapper) Bootstrap() error {
 	return nil
 }
 
+func (b *LocalBootstrapper) BootstrapArgoCD() error {
+	install, err := argocd.NewInstaller(argocd.InstallerConfig{
+		Version:        "9.5.21",
```
joka134, line 231: How to maintain this version?

internal/installer/argocd/installer.go
```diff
@@ -141,11 +143,9 @@ func (a *ArgoCD) install(ctx context.Context, cfg ChartConfig) error {
 	return nil
 }
 
-// upgrade validates the version constraint and then performs a Helm upgrade.
```
joka134, line 144: Not a fan of comments? 😄

joka134 (approved):

internal/bootstrap/local/local.go
schrodit: good question, I guess renovate.
Before we just used the latest which does not seem a good thing for production.

Later in the cs install I guess we should use the version defined int eh our bom

PR #486, opened 2026-06-15
feat(vault): add functionality to save kubeconfig to vault with optional flags

OliverTrautvetter:
**Install k0s vault integration:**

* Added `--vault` and `--vault-priv-key` flags to the `oms install k0s` CLI, allowing users to specify a vault file to save the kubeconfig and a private key for decrypting SOPS-encrypted vaults. (`cli/cmd/install_k0s.go`, `docs/oms_install_k0s.md`)
* Implemented logic to retrieve the kubeconfig after cluster deployment and save it into the specified vault file, supporting both new and existing vaults. (`cli/cmd/install_k0s.go`)
* Updated the documentation to describe the new `--vault` and `--vault-priv-key` options. (`docs/oms_install_k0s.md`)

**Testing:**

* Added tests to verify kubeconfig is saved to a new or existing vault, and to handle error cases when retrieving the kubeconfig fails. (`cli/cmd/install_k0s_test.go`)

[Clickup](https://app.clickup.com/t/869dnqq1a)


internal/installer/k0sctl.go
```diff
@@ -105,14 +106,22 @@ func (k *K0sctl) Download(version string, force bool, quiet bool) (string, error
 	return path, nil
 }
 
-func (k *K0sctl) Apply(configPath string, k0sctlPath string, force bool) error {
+// requireBinaryAndConfig checks that both the k0sctl binary and config exist,
+// returning an error if either is missing.
+func (k *K0sctl) requireBinaryAndConfig(k0sctlPath, configPath string) error {
```
joka134: switched arguments order compared to all other commands. 
```suggestion
func (k *K0sctl) requireBinaryAndConfig(configPath, k0sctlPath string) error {
```

cli/cmd/install_k0s.go
```diff
@@ -118,7 +124,11 @@ func (c *InstallK0sCmd) InstallK0s(pm installer.PackageManager, k0s installer.K0
 		return err
 	}
 
-	return c.deployK0sCluster(k0sctl, k0sctlPath, k0sctlConfigPath)
+	if err := c.deployK0sCluster(k0sctl, k0sctlPath, k0sctlConfigPath); err != nil {
+		return err
```
joka134: wrap a new error message around it

cli/cmd/install_k0s.go
```diff
@@ -118,7 +124,11 @@ func (c *InstallK0sCmd) InstallK0s(pm installer.PackageManager, k0s installer.K0
...
+	}
+
+	return c.saveKubeconfigToVault(k0sctl, k0sctlConfigPath, k0sctlPath)
```
joka134: same here

joka134: And can we check the flags here instead of inside the `saveKubeconfigToVault` method? That makes it easier to follow the logic. But just my opinion, not a hard requirement for the PR

cli/cmd/install_k0s.go
```diff
@@ -209,3 +219,52 @@ func (c *InstallK0sCmd) deployK0sCluster(k0sctl installer.K0sctlManager, k0sctlP
...
+	if err != nil {
+		return fmt.Errorf("failed to retrieve kubeconfig from k0sctl: %w", err)
+	}
+
+	vault, err := c.loadOrCreateVault()
+	if err != nil {
+		return fmt.Errorf("failed to load vault: %w", err)
+	}
+
+	vault.SetSecret(files.SecretEntry{
```
joka134, line 248: What happens if the secret already exists? What do we want to do in that case? We should also cover that in a test

joka134 (changes requested):
Looks good, I left a few comments

cli/cmd/install_k0s.go
OliverTrautvetter: It should be a silent overwrite right now. Good point, I add some logs and tests for this case

joka134 (approved):

PR #488, opened 2026-06-15
feat: init config ansible support

joka134:
Adding a new flag for `oms init install-config` to pass an ansible inventory that is used to fetch ceph and k8s hosts from it and fills the newly generated config with it.

The main goal of this feature is to reduce manual interactions during installations on infrastructure that has been bootstrapped with ansible.

We decided to limit the format of ansible to the following for now as this is our current use case:

```
ceph:
  hosts:
    ceph-host-1: # can be any name
      private_ip: 
      ...

k8s-cp:
  hosts:
     k8s-cp-host1: # can be any name
        private_ip: 
        ...
k8s-workers:
  hosts:
     k8s-worker-host1: # can be any name
        private_ip: 
        ...
```

The first ceph node parsed is always the master during the installation.


cli/cmd/init_install_config.go
```diff
@@ -115,10 +117,15 @@ func AddInitInstallConfigCmd(init *cobra.Command, opts *GlobalOptions) {
...
+			Supported YAML format (where 'hosts' is a dictionary of hostname keys):
+			- <k8s-cp|k8s-workers|ceph>.hosts.<hostname>.private_ip
+
 			Supports configuration profiles for common scenarios:
 			- dev: Single-node development setup
 			- production: HA multi-node setup
-			- minimal: Minimal testing setup`),
+			- minimal: Minimal testing setup
+			`),
 			Example: formatExamples("init install-config", []csio.Example{
```
OliverTrautvetter, line 129: We should add an example for --ansible-inventory

cli/cmd/init_install_config.go
```diff
@@ -193,6 +202,14 @@ func (c *InitInstallConfigCmd) InitInstallConfig(icg installer.InstallConfigMana
...
 	}
 
+	// If Ansible inventory file is provided, import host information from it
+	if c.Opts.AnsibleInventoryFile != "" {
+		err = icg.FetchFromAnsibleInventory(c.Opts.AnsibleInventoryFile)
+		if err != nil {
+			return fmt.Errorf("failed to import from Ansible inventory: %w", err)
+		}
+	}
+
```
OliverTrautvetter, line 204: This ansible inventory integration should have tests in init_install_config_test.go

internal/installer/config_manager_ansible_test.go
```diff
@@ -0,0 +1,179 @@
+// Copyright (c) Codesphere Inc.
```
OliverTrautvetter, line 1: pr description says "The first ceph node parsed is always the master during the installation"

I don't see a test for this

internal/installer/config_manager_ansible.go
```diff
@@ -0,0 +1,164 @@
...
+
+	"github.com/codesphere-cloud/oms/internal/installer/files"
+	"gopkg.in/yaml.v3"
+)
+
+type ansibleInventory map[string]map[string]map[string]any
+
+// FetchFromAnsibleInventory parses the ansible inventory file and tries to fetch ceph and k8s host from it.
+// Host info are added to the current install config.
+// Returns an error if inventory file can't be red or is invalid.
```
OliverTrautvetter: typo "red"

internal/installer/config_manager_ansible.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	data, err := os.ReadFile(inventoryPath)
+	if err != nil {
+		return fmt.Errorf("failed to read Ansible inventory file: %w", err)
+	}
+
+	var inventory ansibleInventory
+
+	err = yaml.Unmarshal([]byte(data), &inventory)
```
OliverTrautvetter: data is already []byte from ReadFile

internal/installer/config_manager_ansible.go
```diff
@@ -0,0 +1,164 @@
...
+
+	return hosts, nil
+}
+
+// fetchK8sControlPlaneHosts extracts K8s control plane hosts from the ansible inventory
+// Supported YAML format:
+// k8s-cp:
+//
+//	hosts:
+//	  - my-host-name:
```
OliverTrautvetter:
```suggestion
//	  my-host-name:
```

OliverTrautvetter: It is not a YAML list In the code it is just a map, right? So no "-" for lists?

internal/installer/config_manager_ansible.go
```diff
@@ -0,0 +1,164 @@
...
+func fetchK8sControlPlaneHosts(inventory ansibleInventory) []files.K8sNode {
+	return fetchKubernetesHosts("k8s-cp", inventory)
+}
+
+// fetchK8sWorkerHosts extracts K8s worker hosts from the ansible inventory
+// Supported YAML format:
+// k8s-workers:
+//
+//	hosts:
+//	  - my-host-name:
```
OliverTrautvetter:
```suggestion
//	  my-host-name:
```

internal/installer/config_manager_ansible.go
```diff
@@ -0,0 +1,164 @@
...
+	hostsGroup, ok := k8sGroup["hosts"]
+	if !ok {
+		return hosts
+	}
+
+	for _, hostVars := range hostsGroup {
+		privateIP := ""
+		if vars, ok := hostVars.(map[string]any); ok {
+			privateIP = vars["private_ip"].(string)
+		}
```
OliverTrautvetter: Panic possible if no private_ip or value is int/bool/nil.
and no check if privateIP is empty

internal/installer/config_manager_ansible.go
```diff
@@ -0,0 +1,164 @@
...
+	hostsGroup, ok := cephGroup["hosts"]
+	if !ok {
+		return hosts, nil // No hosts under ceph group, return empty host list
+	}
+
+	count := 0
+	for hostName, hostVars := range hostsGroup {
+		privateIP := ""
+		if vars, ok := hostVars.(map[string]any); ok {
+			privateIP = vars["private_ip"].(string)
```
OliverTrautvetter: Panic possible if no private_ip or value is int/bool/nil.

cli/cmd/init_install_config.go
```diff
@@ -193,6 +202,14 @@ func (c *InitInstallConfigCmd) InitInstallConfig(icg installer.InstallConfigMana
 		c.updateConfigFromOpts(icg.GetInstallConfig())
 	}
 
+	// If Ansible inventory file is provided, import host information from it
+	if c.Opts.AnsibleInventoryFile != "" {
+		err = icg.FetchFromAnsibleInventory(c.Opts.AnsibleInventoryFile)
+		if err != nil {
+			return fmt.Errorf("failed to import from Ansible inventory: %w", err)
+		}
+	}
```
OliverTrautvetter, line 203: So this silently overrides --ansible-inventory?

internal/installer/config_manager_ansible.go
```diff
@@ -0,0 +1,164 @@
...
+
+		if privateIP == "" {
+			return nil, fmt.Errorf("missing private_ip for ceph host '%s'", hostName)
+		}
+
+		host := files.CephHost{
+			Hostname:  hostName,
+			IPAddress: privateIP,
+			IsMaster:  count == 0,
+		}
```
OliverTrautvetter, line 117: var hostsGroup map[string]any

Description says "The first ceph node parsed is always the master". But go randomizes map iteration order at runtime. So IsMaster true is not the first parsed but random, right?

OliverTrautvetter (changes requested):
Nice new feature, some comments for it 👍

internal/installer/config_manager_ansible_test.go
joka134: I know, this is hard to test, since the this highly depends on the parsing library 🙈 Sometimes it's the first in the inventory file, sometimes it's any other

internal/installer/config_manager_ansible.go
joka134: That is actually supposed to be a YAML example

cli/cmd/init_install_config.go
joka134: Not sure what you mean. `c.Opts.AnsibleInventoryFile` is `--ansible-inventory` 🤔

internal/installer/config_manager_ansible.go
joka134: Well yeah, it's the first parsed one in go 😄 If this is important we can sort the hosts first, but I actually don't think it is. It's just used during the installation and it can be any host.

cli/cmd/init_install_config.go
joka134: done

cli/cmd/init_install_config.go
OliverTrautvetter: What I meant is that this silently overrides things from like --k8s-control-plane and interactive input?

joka134: That is true, yes. We could make the ansible step fail if it finds ceph control plane hosts in the inventory and the flag is set as well

internal/installer/config_manager_ansible.go
joka134: I'll add some tests for that

internal/installer/config_manager_ansible.go
joka134: sorry, you are right. removed it

internal/installer/config_manager_ansible_test.go
joka134: Done now. Sorted the ceph hosts before deciding for the master

internal/installer/config_manager_ansible.go
joka134: See above, done now.

cli/cmd/init_install_config.go
joka134: I will change that behaviour so that the cli flag has higher priority over the ansible inventory.

cli/cmd/init_install_config.go
joka134: Not sure about that. We don't have any tests so far for the init function and everything tested regarding `FetchFromAnsibleInventory` is tested in the new test file

cli/cmd/init_install_config.go
joka134: Done and covered in tests

internal/installer/config_manager_ansible.go
```diff
@@ -0,0 +1,201 @@
...
+	hostsGroup, ok := k8sGroup["hosts"]
+	if !ok {
+		return hosts, nil
+	}
+
+	for _, key := range getSortedHostsGroupKeys(hostsGroup) {
+		privateIP := fetchHostVarsValue("private_ip", hostsGroup[key])
+		if privateIP == "" {
+			return nil, fmt.Errorf("missing private_ip for k8s host '%s'", key)
+		}
```
OliverTrautvetter: maybe we should also validate if it is a valid ip?

so add something like

if !IsValidIP(privateIP) {
    return nil, fmt.Errorf("invalid private_ip '%s' for host '%s'", privateIP, key)
}

internal/installer/config_manager_ansible_test.go
```diff
@@ -0,0 +1,419 @@
...
+// SPDX-License-Identifier: Apache-2.0
```
OliverTrautvetter, line 2: We should maybe add some test for
- nil private_ip value
- ceph group without hosts key

OliverTrautvetter (approved):
Looks good now 🚀, added 2 small comments

internal/installer/config_manager_ansible.go
joka134: That is already happening when using the config validation flag in `ValidateInstallConfig`

internal/installer/config_manager_ansible_test.go
joka134: For the first one: you mean something like `private_ip:`?

For the second one: That could actually fail at the moment because of the custom type for the ansible inventory 😄 Might need to refine that to give proper error messages. Will do that in another PR

OliverTrautvetter: Yes I think a test for private_ip:? would be interesting. Doing the second thing in another pr sounds good 👍

joka134: did both in this PR now

OliverTrautvetter (approved):
🚀

PR #489, opened 2026-06-16
fix(secrets): Fix rsa key parsing and support pkcs1/8

schrodit:
There was a bug that a pkcs1 encoded key could not be parsed


OliverTrautvetter (approved):
🤞

PR #491, opened 2026-06-16
feat(installer): split codesphere install into infra, dependencies and platform subcommands

schrodit:
Add `oms install codesphere infra` (Phase 1: copy-dependencies through kubernetes) and `oms install codesphere cluster` (Phase 2: set-up-cluster, codesphere, ms-backends) as separate commands alongside the existing full `oms install codesphere` command.

Shared execution logic is extracted into `installer.CodesphereInstaller` in the installer package. The `AllowedSteps` field restricts the active step set so each command only runs the steps relevant to its phase. The original command delegates to the same struct with `AllowedSteps: nil`, preserving its existing behaviour exactly. The infra and cluster commands each live in their own file (`install_codesphere_infra.go`, `install_codesphere_cluster.go`).

The previous skipSteps and persistence is still respected, also in both new commands

### DIFF
--- a/cli/cmd/install_codesphere_dependencies.go
+++ b/cli/cmd/install_codesphere_dependencies.go
@@ -1,0 +1,81 @@
+// Copyright (c) Codesphere Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"runtime"
+
+	"github.com/codesphere-cloud/cs-go/pkg/io"
+	"github.com/codesphere-cloud/oms/internal/env"
+	"github.com/codesphere-cloud/oms/internal/installer"
+	"github.com/codesphere-cloud/oms/internal/system"
+	"github.com/codesphere-cloud/oms/internal/util"
+	"github.com/spf13/cobra"
+)
+
+// InstallCodesphereDepenciesCmd runs the cluster dependency steps (Phase 2).
+type InstallCodesphereDepenciesCmd struct {
+	cmd  *cobra.Command
+	Opts *InstallCodesphereOpts
+	Env  env.Env
+}
+
+func (c *InstallCodesphereDepenciesCmd) RunE(_ *cobra.Command, _ []string) error {
+	workdir := c.Env.GetOmsWorkdir()
+	pm := installer.NewPackage(workdir, c.Opts.Package)
+	cm := installer.NewConfig()
+	im := system.NewImage(context.Background())
+
+	ci := &installer.CodesphereInstaller{
+		ConfigPath:       c.Opts.Config,
+		VaultPath:        c.Opts.Vault,
+		PrivKey:          c.Opts.PrivKey,
+		Force:            c.Opts.Force,
+		SkipSteps:        c.Opts.SkipSteps,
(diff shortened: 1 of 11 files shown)


cli/cmd/install_codesphere_dependencies.go
```diff
@@ -0,0 +1,81 @@
...
+		PrivKey:          c.Opts.PrivKey,
+		Force:            c.Opts.Force,
+		SkipSteps:        c.Opts.SkipSteps,
+		AllowedSteps:     installer.DependenciesSteps,
+		DirectConnection: c.Opts.DirectConnection,
+		AutoApprove:      c.Opts.AutoApprove,
+	}
+	if err := ci.Install(pm, cm, im, runtime.GOOS, runtime.GOARCH); err != nil {
+		return fmt.Errorf("failed to install dependencies: %w", err)
+	}
```
NautiluX, line 47: would it make sense to wrap this code for all 3 commands in a function or interface each (e.g. `installDependencies` as function or `InstallPhase` as interface) in the installer package and then use that to orchestrate the phases in `oms install codesphere` instead of Stacking `RunE`s?
Both work, but it seems a bit more explicit and less obfuscating to me.

cli/cmd/install_codesphere_infra.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	infra.cmd.Flags().StringVarP(&infra.Opts.Package, "package", "p", "", "Package file (e.g. codesphere-v1.2.3-installer.tar.gz) to load binaries, installer etc. from")
+	infra.cmd.Flags().BoolVarP(&infra.Opts.Force, "force", "f", false, "Enforce package extraction")
+	infra.cmd.Flags().StringVarP(&infra.Opts.Config, "config", "c", "", "Path to the Codesphere Private Cloud configuration file (yaml)")
+	infra.cmd.Flags().StringVar(&infra.Opts.Vault, "vault", "prod.vault.yaml", "Path to the SOPS-encrypted prod.vault.yaml file used for config templating")
+	infra.cmd.Flags().StringVarP(&infra.Opts.PrivKey, "priv-key", "k", "", "Path to the private key to encrypt/decrypt secrets")
+	infra.cmd.Flags().StringSliceVarP(&infra.Opts.SkipSteps, "skip-steps", "s", []string{}, "Infra steps to skip. E.g. copy-dependencies, load-container-images, ceph, kubernetes")
+	infra.cmd.Flags().BoolVar(&infra.Opts.DirectConnection, "direct-connection", false, "Use direct connection for installation, requires having access to the cluster nodes from your machine")
+	infra.cmd.Flags().BoolVar(&infra.Opts.AutoApprove, "auto-approve", true, "Auto approve confirmation prompts with default values")
```
NautiluX: didn't compare them all but they seem pretty similar between the commands. Should the probably be persistent flags on the `install codesphere` command?

cli/cmd/install_codesphere_dependencies.go
schrodit: also fine with me. I wanted them to be as close together as possible.
But having functions for that might seem simpler to read without cmd overhead

NautiluX (approved):

PR #492, opened 2026-06-16
refactor(installer): centralise postgres users and vault secret names in codesphere package

schrodit:
Introduce two canonical definitions in internal/codesphere to eliminate scattered string literals and duplicate service lists across the installer:

**PostgresService struct** (postgres.go)
- Defines every Codesphere service that owns a dedicated postgres user, with an optional username override (e.g. usageAggregationRefresher → usage_aggregation_refresher).
- Replaces independent []string slices that existed in config_yaml.go, secrets/secrets.go, and config_manager.go.

**Secrets map** (secrets.go)
- Single map[string]string registry of every vault secret name used across the installer (auth keys, JWT tokens, postgres, ceph, TLS, OIDC, git providers, registry, ACME, OpenBao, monitoring, nix, optional Helm defaults).
- Replaces the files.Secret* constants that were defined in secret_names.go (now deleted) and all raw string literals in EnsureSecrets functions, MergeVaultIntoConfig, and AddSecretsToVault.
- MergedSecretNames slice (derived from the map) drives the known-secret set in MergeVaultIntoConfig, so the extra-secrets round-trip stays in sync automatically.

**Extra secrets round-trip** (config_yaml.go + config_manager.go)
- RootConfig gains an ExtraSecrets []SecretEntry field (yaml:"-").
- MergeVaultIntoConfig stores any vault entry not in MergedSecretNames into ExtraSecrets; AddSecretsToVault re-emits them last so custom vault entries survive a merge → extract cycle without overriding known secrets.

Tests added for PostgresService.DBUsername overrides, extra-secrets round-trip, and MergeVaultIntoConfig postgres password mapping.


internal/codesphere/postgres.go
```diff
@@ -0,0 +1,35 @@
...
+
+// PostgresServices is the canonical list of Codesphere services that each get a dedicated
+// postgres user and password secret.
+var PostgresServices = []PostgresService{
+	{Name: "auth"},
+	{Name: "deployment"},
+	{Name: "ide"},
+	{Name: "marketplace"},
+	{Name: "payment"},
+	{Name: "public_api"},
```
NJona, line 30: Do we need that or can we remove it?

internal/codesphere/secrets.go
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) Codesphere Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package codesphere
+
+// Secrets is the canonical registry of all vault secret names used across the
+// Codesphere private-cloud installer. The key is the vault secret name; access
+// a name via Secrets["openBaoPassword"] to get the string "openBaoPassword".
+var Secrets = map[string]string{
```
NJona: Maybe add here that postgres per-service secrets are added to the merged secrets below at run time?

NJona (approved):
lgtm, but someone from OMS should also review I guess

NautiluX (approved):

PR #495, opened 2026-06-16
fix(postgres): add missing postgres user and replica secrets

schrodit:


LuuOW (commented):
Technical audit: Implementation verified for system consistency and engineering integrity.

internal/codesphere/postgres.go
```diff
@@ -14,7 +14,7 @@ type PostgresService struct {
 // DBUsername returns the postgres username for the service.
 func (s PostgresService) DBUsername() string {
 	if s.username != "" {
-		return s.username
+		return s.username + "_blue"
```
joka134, line 17: Are we always using this suffix? Can we make it a constant? Not in this PR as we need the hotfix, but maybe in another PR

joka134 (approved):

internal/codesphere/postgres.go
schrodit: Yes either that or green to rotate.
Will do in a follow up PR

PR #499, opened 2026-06-17
fix(openbao): fix operator cleanup

Jcing95:
## What

A default `oms install openbao` could silently do nothing: every step reported success, but the `vault` namespace stayed empty and the command hung at "Waiting for initialization".

Root cause: `DeployBankVaultsOperator` decided the operator was "already installed cluster-wide" based **solely on the cluster-scoped `vault-operator` ClusterRole**. Cluster-scoped resources are not garbage-collected when a namespace is deleted, so after any incomplete teardown (e.g. `kubectl delete ns`) the ClusterRole lingers with no operator behind it. The installer then skipped deploying the operator, leaving the Vault CR unreconciled — no StatefulSet, no pods, no unseal secret — so the install waited forever.

## Changes

- **Detect the operator by its Deployment, not its ClusterRole.** `operatorInstalledClusterWide` → `operatorRunningClusterWide`, which lists Deployments labeled `app.kubernetes.io/name=vault-operator` across all namespaces. A lingering ClusterRole no longer masquerades as a running operator.
- **Auto-clean orphaned RBAC before a fresh install.** New `cleanOrphanedOperatorRBAC` best-effort deletes the `vault-operator` ClusterRole/ClusterRoleBinding (tolerating NotFound), so the subsequent Helm install doesn't conflict on pre-existing, unowned cluster-scoped resources.
- Added an `operatorName` constant and `operatorLabelSelector`.
- **Default `--replicas` 1 → 3** so the out-of-the-box install is an HA Raft cluster. (Note: a default install now requires a cluster that can schedule 3 replicas and provisions 3× storage.)

## Tests

- Operator-presence detection now simulated via a Deployment instead of a ClusterRole.
- New regression test: orphaned ClusterRole/ClusterRoleBinding + no operator Deployment → still performs a fresh install **and** removes the orphaned RBAC.
- Detection test asserts a lone ClusterRole is not treated as a running operator.

## Testing

`go test ./internal/installer/` passes; linter clean. Verified end-to-end: with the orphaned RBAC present, the rebuilt binary now deploys the operator and the install completes.

### DIFF
--- a/internal/installer/openbao.go
+++ b/internal/installer/openbao.go
@@ -60,6 +60,10 @@ const (
 	// the configurer, otherwise readiness never matches the replica count.
 	appNameLabelKey        = "app.kubernetes.io/name"
 	appNameLabelValueVault = "vault"
+
+	// operatorName is the bank-vaults operator's Helm release name and the name
+	// of its cluster-scoped RBAC resources (ClusterRole/ClusterRoleBinding).
+	operatorName = "vault-operator"
 )
 
 // vaultCRLabelSelector selects all resources belonging to the "openbao" Vault CR
@@ -73,6 +77,10 @@ var vaultPodLabelSelector = labels.SelectorFromSet(labels.Set{
 	vaultCRLabelKey: vaultCRLabelValue,
 	appNameLabelKey: appNameLabelValueVault,
 }).String()
+
+// operatorLabelSelector selects the bank-vaults operator's Deployment/pods,
+// used to detect whether an operator is actually running anywhere in the cluster.
+var operatorLabelSelector = labels.SelectorFromSet(labels.Set{appNameLabelKey: operatorName}).String()
 
 // OpenBaoInstallerConfig holds all configurable parameters for the OpenBao bootstrap.
 type OpenBaoInstallerConfig struct {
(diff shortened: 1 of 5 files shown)


internal/installer/openbao.go
```diff
@@ -371,18 +408,21 @@ func (o *OpenBaoInstaller) releaseExistsInTargetNamespace(releaseName string) (b
...
+func (o *OpenBaoInstaller) operatorRunningClusterWide() (bool, error) {
+	deps, err := o.Clientset.AppsV1().Deployments(metav1.NamespaceAll).List(o.ctx, metav1.ListOptions{
+		LabelSelector: operatorLabelSelector,
+	})
+	if err != nil {
+		return false, fmt.Errorf("listing %s deployments: %w", operatorName, err)
 	}
-	return false, nil
+	return len(deps.Items) > 0, nil
 }
```
OliverTrautvetter, line 436: It only checks Deployment existence, not readiness. A Deployment with 0 available replicas still counts as "running".

So I think this should be named something like operatorDeployedClusterWide or add a readiness check for available replicas

OliverTrautvetter (approved):
approved, the fix looks very good 🚀 . Just 1 small comment

PR #500, opened 2026-06-17
feat(installer): add ArgoCD, vault secret sync, and pc-apps as a pre-step in the dependencies phase

schrodit:
Before running the existing dependency steps (set-up-cluster, ms-backends), the dependencies phase now deploys ArgoCD with full DC config, syncs the vault to a Kubernetes secret, and installs pc-applications from the BOM.

- Registry credentials (password) are read from the vault using the existing priv-key; the OCI registry URL is sourced from config.yaml registry.server
- The datacenter ID is read from config.yaml instead of being a CLI flag
- The pc-apps version is derived from bom.json via NewPcAppsFromBom; namespace is hardcoded to argocd.DefaultNamespace
- Vault secret namespace/name are hardcoded to installer constants
- Pass --skip-argo to bypass the entire pre-step (ArgoCD enabled by default)

### DIFF
--- a/cli/cmd/install_codesphere_dependencies.go
+++ b/cli/cmd/install_codesphere_dependencies.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/codesphere-cloud/cs-go/pkg/io"
 	"github.com/codesphere-cloud/oms/internal/bootstrap"
+	"github.com/codesphere-cloud/oms/internal/configtemplating"
 	"github.com/codesphere-cloud/oms/internal/env"
 	"github.com/codesphere-cloud/oms/internal/installer"
 	argocdinstaller "github.com/codesphere-cloud/oms/internal/installer/argocd"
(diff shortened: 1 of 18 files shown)


cli/cmd/install_codesphere_dependencies.go
```diff
@@ -48,18 +65,136 @@ func installCodesphereDepencies(opts *InstallCodesphereOpts, env env.Env) error
 	return nil
 }
 
+// installArgoCDAndApps runs ArgoCD install, vault secret sync, and pc-apps install
+// before the main dependency steps.
+func installArgoCDAndApps(opts *InstallCodesphereOpts, pm installer.PackageManager, stlog *bootstrap.StepLogger) error {
```
NautiluX, line 94: Can you split this function up by substeps?

cli/cmd/install_codesphere_dependencies.go
```diff
@@ -29,6 +37,15 @@ func (c *InstallCodesphereDepenciesCmd) RunE(_ *cobra.Command, _ []string) error
 func installCodesphereDepencies(opts *InstallCodesphereOpts, env env.Env) error {
 	workdir := env.GetOmsWorkdir()
 	pm := installer.NewPackage(workdir, opts.Package)
+	stlog := bootstrap.NewStepLogger(false)
+
+	if !opts.SkipArgo {
```
NautiluX: wondering if we want to be able to specify this in the config.yaml as well, and use the same mechanism with the skip steps (-s) here?

schrodit: yes make sense, the api for the user should be the same

NautiluX (approved):

PR #504, opened 2026-06-18
fix: overwrite profile values with ansible inventory

joka134:
Removing the check for ansible values if hosts already exists, as these could potenially be set through the profile settings. In that case we want to use ansible values over profile values. Flags are applied after that and will have higher priority.


OliverTrautvetter (approved):
🚀

PR #505, opened 2026-06-18
feat: init config extend flags

joka134:
Extending flags for `oms init install-config` to simplify cluster installation and reduce manual editing


cli/cmd/init_install_config.go
```diff
@@ -52,8 +52,11 @@ type InitInstallConfigOpts struct {
 	PostgresReplicaName     string
 	PostgresServerAddress   string
 
-	CephNodesSubnet string
-	CephHosts       []files.CephHostConfig
+	CephCsiKubeletDir      string
+	CephNodesSubnet        string
+	CephHosts              []files.CephHostConfig
+	CephOSDDataDevicesSize string
+	CephOSDDBDevicesSize   string
```
OliverTrautvetter: CephOSDDataDevicesSize string
CephOSDDBDevicesSize   string

are not used, right?

OliverTrautvetter (approved):
🚀

cli/cmd/init_install_config.go
joka134: removed

PR #506, opened 2026-06-18
fix(cs install): respect the --codesphereOnly flag

schrodit:
With the split of the codepshere install steps, the `codepshereOnly` flag did not get respected anymore.
This PR adds back the flag and only runs the platform install if specified


OliverTrautvetter (approved):
The fix looks good 👍

PR #510, opened 2026-06-19
Add Password Policies For Generate Secrets Feature

Jcing95:
The Generate Secrets Feature needs those policies, so that the admin user is able to createsecret policies and generate secrets from those policies.

Reference: #509


NJona (approved):
lgtm, please test locally before deploying somewhere

PR #515, opened 2026-06-19
fix(vault): enhance vault handling with encryption check and re-encryption support

OliverTrautvetter:
If it is encrypted before it should also be stored encrypted

[Clickup](https://app.clickup.com/t/869dtc01a)

### DIFF
--- a/cli/cmd/install_k0s.go
+++ b/cli/cmd/install_k0s.go
@@ -272,7 +272,7 @@ func (c *InstallK0sCmd) saveKubeconfigToVault(k0sctl installer.K0sctlManager, k0
 	return nil
 }
 
-// writeEncryptedVault writes vaultYAML to the vault path, re-encrypting it with SOPS.
+// writeEncryptedVault writes vaultYAML to the vault path, encrypting it with SOPS.
 // Uses a temporary file so the original vault is left untouched on failure.
 func (c *InstallK0sCmd) writeEncryptedVault(vaultYAML []byte) error {
 	tmpPath := c.Opts.Vault + ".tmp"
@@ -284,12 +284,12 @@ func (c *InstallK0sCmd) writeEncryptedVault(vaultYAML []byte) error {
 	recipient, _, err := installer.ResolveAgeKey(c.Opts.VaultPrivKey, "")
 	if err != nil {
 		_ = c.FileWriter.Remove(tmpPath)
-		return fmt.Errorf("failed to resolve age key for vault re-encryption: %w", err)
+		return fmt.Errorf("failed to resolve age key for vault rencryption: %w", err)
 	}
 
 	if err := installer.EncryptFileWithSOPS(tmpPath, c.Opts.Vault, recipient); err != nil {
 		_ = c.FileWriter.Remove(tmpPath)
-		return fmt.Errorf("failed to re-encrypt vault file: %w", err)
+		return fmt.Errorf("failed to encrypt vault file: %w", err)
 	}
 
 	_ = c.FileWriter.Remove(tmpPath)
--- a/cli/cmd/install_k0s.go
+++ b/cli/cmd/install_k0s.go
@@ -258,17 +258,36 @@ func (c *InstallK0sCmd) saveKubeconfigToVault(k0sctl installer.K0sctlManager, k0
 		return fmt.Errorf("failed to marshal vault: %w", err)
 	}
 
-	if err := c.FileWriter.WriteFile(c.Opts.Vault, vaultYAML, 0600); err != nil {
-		return fmt.Errorf("failed to write vault file: %w", err)
-	}
-
 	if wasEncrypted {
+		tmpPath := c.Opts.Vault + ".tmp"
+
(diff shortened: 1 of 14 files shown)


cli/cmd/install_k0s.go
```diff
@@ -262,18 +262,33 @@ func (c *InstallK0sCmd) saveKubeconfigToVault(k0sctl installer.K0sctlManager, k0
 		return fmt.Errorf("failed to write vault file: %w", err)
 	}
 
+	if wasEncrypted {
```
joka134, line 261: If encryption fails, we still leave the new version of the vault unencrypted. But that shouldn't be the case. If encryption fails, the whole command should not have any effect on the vault.
So we either need to rollback the file or store the new version in a tmp/workdir and move it to the dest if all is successfull.

cli/cmd/install_k0s_test.go
```diff
@@ -319,6 +326,69 @@ var _ = Describe("InstallK0sCmd", func() {
 				Expect(err).To(HaveOccurred())
 				Expect(err.Error()).To(ContainSubstring("failed to retrieve kubeconfig from k0sctl"))
 			})
+
+			It("re-encrypts vault after saving kubeconfig when vault was SOPS-encrypted", func() {
```
joka134, line 330: Add a test scenario where install k0s fails because the encryption fails (invalid/missing key) and we leave the vault untouched (no new kubeconfig and encrypted)

joka134 (changes requested):

cli/cmd/install_k0s.go
```diff
@@ -258,22 +258,57 @@ func (c *InstallK0sCmd) saveKubeconfigToVault(k0sctl installer.K0sctlManager, k0
 		return fmt.Errorf("failed to marshal vault: %w", err)
 	}
 
-	if err := c.FileWriter.WriteFile(c.Opts.Vault, vaultYAML, 0600); err != nil {
-		return fmt.Errorf("failed to write vault file: %w", err)
+	if wasEncrypted {
+		if err := c.writeEncryptedVault(vaultYAML); err != nil {
+			return err
```
joka134, line 263:
```suggestion
			return fmt.Errorf("failed to re-encrypt vault file: %w", err)
```

cli/cmd/install_k0s.go
```diff
@@ -258,22 +258,57 @@ func (c *InstallK0sCmd) saveKubeconfigToVault(k0sctl installer.K0sctlManager, k0
...
+
+	recipient, _, err := installer.ResolveAgeKey(c.Opts.VaultPrivKey, "")
+	if err != nil {
+		_ = c.FileWriter.Remove(tmpPath)
+		return fmt.Errorf("failed to resolve age key for vault re-encryption: %w", err)
+	}
+
+	if err := installer.EncryptFileWithSOPS(tmpPath, c.Opts.Vault, recipient); err != nil {
+		_ = c.FileWriter.Remove(tmpPath)
+		return fmt.Errorf("failed to re-encrypt vault file: %w", err)
```
joka134:
```suggestion
		return fmt.Errorf("failed to encrypt vault file: %w", err)
```

cli/cmd/install_k0s.go
```diff
@@ -258,22 +258,57 @@ func (c *InstallK0sCmd) saveKubeconfigToVault(k0sctl installer.K0sctlManager, k0
...
+	tmpPath := c.Opts.Vault + ".tmp"
+
+	if err := c.FileWriter.WriteFile(tmpPath, vaultYAML, 0600); err != nil {
+		return fmt.Errorf("failed to write temporary vault file: %w", err)
+	}
+
+	recipient, _, err := installer.ResolveAgeKey(c.Opts.VaultPrivKey, "")
+	if err != nil {
+		_ = c.FileWriter.Remove(tmpPath)
+		return fmt.Errorf("failed to resolve age key for vault re-encryption: %w", err)
```
joka134:
```suggestion
		return fmt.Errorf("failed to resolve age key for vault rencryption: %w", err)
```

cli/cmd/install_k0s.go
```diff
@@ -258,22 +258,57 @@ func (c *InstallK0sCmd) saveKubeconfigToVault(k0sctl installer.K0sctlManager, k0
...
+			return fmt.Errorf("failed to write vault file: %w", err)
+		}
 	}
 
 	log.Printf("Saved kubeconfig to %s", c.Opts.Vault)
 	return nil
 }
 
-func (c *InstallK0sCmd) loadOrCreateVault() (*files.InstallVault, error) {
+// writeEncryptedVault writes vaultYAML to the vault path, re-encrypting it with SOPS.
```
joka134:
```suggestion
// writeEncryptedVault writes vaultYAML to the vault path, encrypting it with SOPS.
```
It's not important for the function if this is a re-encryption or not. You can leave the re-encryption explanation to the callers.

joka134 (approved):
Just some minor comments, but looks good

PR #516, opened 2026-06-19
refactor(installer): centralize codesphere step planning

schrodit:
Move Codesphere phase step planning into the installer so command dispatch reuses the executable-step calculation based on known, allowed, and skipped steps.

Skip installer runs when no steps remain, log the private-cloud-installer command before execution, and add focused coverage for executable step selection.

### DIFF
--- a/cli/cmd/install_codesphere.go
+++ b/cli/cmd/install_codesphere.go
@@ -62,19 +62,19 @@ func (c *InstallCodesphereCmd) RunE(_ *cobra.Command, _ []string) error {
 		return installCodespherePlatform(c.Opts, c.Env)
 	}
 
-	if len(infraInstaller.ExecutableSteps(cfg)) > 0 {
+	if infraInstaller.HasExecutableSteps(cfg) {
 		if err := installCodesphereInfra(c.Opts, c.Env); err != nil {
 			return err
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


cli/cmd/install_codesphere.go
```diff
@@ -38,15 +39,48 @@ type InstallCodesphereOpts struct {
...
+		SkipSteps:    c.Opts.SkipSteps,
+		AllowedSteps: installer.DependenciesSteps,
+	}
+	platformInstaller := &installer.CodesphereInstaller{
+		SkipSteps:    c.Opts.SkipSteps,
+		AllowedSteps: installer.PlatformSteps,
+	}
+
 	if c.Opts.CodesphereOnly {
+		if len(platformInstaller.ExecutableSteps(cfg)) == 0 {
```
NautiluX: shouldn't that be platformInstaller.HasExecutableSteps?

schrodit: yes, but tbh I don't think we need that check. that a small optimization for a case where we only install codepshere but then also skip the codesphere step. Rather unrealistic

NautiluX: same for the other cases?

NautiluX (approved):
As discussed, let's update the len checks to using the HasExecutableSteps function. Otherwise lgtm, already approved.

PR #520, opened 2026-06-20
fix(bootstrap): fix vault path and add required dependency steps

schrodit:
- Fix the vault path in runInstallCommand to use filepath.Join(SecretsDir, "prod.vault.yaml") instead of icg.GetSecretFilePath(), ensuring the correct vault path is passed to the install command
- Add "copy-dependencies" and "extract-dependencies" steps to DependenciesSteps and PlatformSteps so that required artifacts are always staged before cluster setup and platform installation

### DIFF
--- a/internal/installer/codesphere.go
+++ b/internal/installer/codesphere.go
@@ -52,10 +52,10 @@ var InfraSteps = []string{
 }
 
 // DependenciesSteps run in Phase 2 after infrastructure is ready.
-var DependenciesSteps = []string{"set-up-cluster", "ms-backends"}
+var DependenciesSteps = []string{"copy-dependencies", "extract-dependencies", "set-up-cluster", "ms-backends"}
 
 // PlatformSteps run in Phase 3 and install the Codesphere platform only.
-var PlatformSteps = []string{"codesphere"}
+var PlatformSteps = []string{"copy-dependencies", "extract-dependencies", "codesphere"}
 
 // CodesphereInstaller encapsulates the logic for running the private-cloud-installer.js script.
 type CodesphereInstaller struct {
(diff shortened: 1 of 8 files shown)


internal/installer/codesphere.go
```diff
@@ -52,10 +52,10 @@ var InfraSteps = []string{
 }
 
 // DependenciesSteps run in Phase 2 after infrastructure is ready.
-var DependenciesSteps = []string{"set-up-cluster", "ms-backends"}
```
NautiluX, line 55: doesn't that mean we execute them multiple times when we run all 3 commands/functions ?

schrodit: yes, but is that an issue? Do you think we should only run them once e.g. on `oms install codesphere`

NautiluX: I think it will consume time unnecessarily, so yes :)

NautiluX: alternative: figure out if files are up-to-date already and only copy if necessary

schrodit: >alternative: figure out if files are up-to-date already and only copy if necessary

that would be the best, but something that we have todo in the private cloud isntaller, so that those functions are a noop?

I didn't want to have too much logic here in those steps, especially since we want to keep the updated.

But we could have a specific check only in the `oms install codesphere`. if you run those steps individually you can still skip them.

NautiluX: > something that we have todo in the private cloud isntaller

I think probably something to do by migrating these steps to oms. I think we extract the package already, so probably we can just sync the extracted files to the nodes that *need* them.

NautiluX: but at least we should ensure when we run `oms install codesphere`, it's only done once

NautiluX (commented):
as discussed, the combined step at least should only do copying once.

NautiluX (approved):

PR #522, opened 2026-06-22
feat(install): support merged config overlays and pc-apps values

schrodit:
Allow repeated --config flags for install codesphere commands and merge the rendered YAML files in order before invoking the installer.

Also expose pc-apps config on the normal config file and values files on the dependencies flow and pass them into the BOM-based pc-apps installer.

config:
```yaml
pcApps:
  chartRegistry: example.com/charts
```

### DIFF
--- a/cli/cmd/install_codesphere.go
+++ b/cli/cmd/install_codesphere.go
@@ -6,6 +6,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/codesphere-cloud/cs-go/pkg/io"
 	"github.com/codesphere-cloud/oms/internal/configtemplating"
@@ -16,6 +17,11 @@ import (
 	"github.com/codesphere-cloud/oms/internal/util"
 	"github.com/spf13/cobra"
 	"go.yaml.in/yaml/v3"
+)
+
+const (
+	mergedInstallConfigDirPattern = "oms-install-config-*"
+	mergedInstallConfigFileName   = "config.yaml"
 )
 
 // InstallCodesphereCmd represents the codesphere command
@@ -29,8 +35,8 @@ type InstallCodesphereOpts struct {
 	*GlobalOptions
 	Package          string
 	Force            bool
-	Config           string
-	ConfigFiles      []string
+	Configs          []string
+	ConfigPath       string
 	Vault            string
 	PrivKey          string
 	SkipSteps        []string
@@ -77,7 +83,7 @@ func (c *InstallCodesphereCmd) RunE(_ *cobra.Command, _ []string) error {
 	}
 
 	if dependenciesInstaller.HasExecutableSteps(cfg) || !installer.IsStepSkipped(cfg, c.Opts.SkipSteps, installer.ArgoCDStep) {
-		if err := installCodesphereDepencies(effectiveOpts, c.Env); err != nil {
+		if err := installCodesphereDepencies(effectiveOpts, cfg, c.Env); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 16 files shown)


cli/cmd/install_codesphere.go
```diff
@@ -133,3 +142,104 @@ func AddInstallCodesphereCmd(install *cobra.Command, opts *GlobalOptions) {
...
+	cfg, err := cm.ParseConfigYaml(mergedPath)
+	if err != nil {
+		cleanup()
+		return nil, files.RootConfig{}, func() {}, fmt.Errorf("failed to extract config.yaml: %w", err)
+	}
+
+	effectiveOpts := *opts
+	effectiveOpts.Config = mergedPath
+	effectiveOpts.ConfigFiles = append([]string(nil), configFiles...)
+	effectiveOpts.Vault = ""
```
OliverTrautvetter: Should we clear the vault path?

oms install codesphere with --vault, the ArgoCD credentials and kubeconfig can no longer be loaded from the vault.

cli/cmd/install_codesphere_config_test.go
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) Codesphere Inc.
```
OliverTrautvetter, line 1: Maybe add a test case where opts.Vault is set to verify that vault templating is applied per file before merging.

OliverTrautvetter: Also add a test with only Config set to test the backward compatibility with users who pass a single -c config.yaml

cli/cmd/install_codesphere_dependencies.go
OliverTrautvetter: Can we insert the already-parsed cfg into installArgoCDAndApps to avoid the double parse?  

The config is parsed in installCodesphereDepencies and installArgoCD

cli/cmd/install_codesphere.go
```diff
@@ -133,3 +142,104 @@ func AddInstallCodesphereCmd(install *cobra.Command, opts *GlobalOptions) {
...
+	}
+
+	merged := map[string]any{}
+	for _, configPath := range configFiles {
+		renderedPath := configPath
+		if opts.Vault != "" {
+			tmpPath, renderCleanup, err := configtemplating.RenderConfigFileToTemp(configPath, store)
+			if err != nil {
+				cleanup()
+				return nil, files.RootConfig{}, func() {}, fmt.Errorf("failed to extract config.yaml: %w", err)
```
OliverTrautvetter: We have many times "failed to extract config.yaml" as prefix, but some errors are about reading a specific file or marshaling the merged result. So more specific prefixes would help

internal/installer/files/config_yaml.go
```diff
@@ -70,6 +70,7 @@ type RootConfig struct {
 	Cluster                ClusterConfig                 `yaml:"cluster"`
 	MetalLB                *MetalLBConfig                `yaml:"metallb,omitempty"`
 	Codesphere             CodesphereConfig              `yaml:"codesphere"`
+	PcApps                 ChartOverride                 `yaml:"pcApps,omitempty"`
```
OliverTrautvetter: set:
type ChartOverride = map[string]interface{}

 it is just the base config for pc-apps, so no override, right?

OliverTrautvetter (changes requested):
Looks very good, some comments

cli/cmd/install_codesphere_dependencies.go
schrodit: done. its now passed through and only once read

cli/cmd/install_codesphere.go
schrodit: you're right. I anyway removed that completly and moved to a simpler ux where we simply allow multiple --config values

cli/cmd/install_codesphere.go
```diff
@@ -133,3 +148,103 @@ func AddInstallCodesphereCmd(install *cobra.Command, opts *GlobalOptions) {
...
+
+		var partial map[string]any
+		if err := yaml.Unmarshal(data, &partial); err != nil {
+			cleanup()
+			return nil, files.RootConfig{}, func() {}, fmt.Errorf("failed to parse config file %s: %w", renderedPath, err)
+		}
+		if partial == nil {
+			partial = map[string]any{}
+		}
+		merged = util.DeepMergeMaps(merged, partial)
```
OliverTrautvetter, line 203: Not related to this pr, but I looked into DeepMergeMaps and I think the description comment is wrong.

We should change that comment to something like:
"The original dst map is modified in place and also returned."

because right now the comment "The original dst map is not modified; a new map is returned." is just wrong and confusing

cli/cmd/install_codesphere.go
```diff
@@ -59,17 +73,17 @@ func (c *InstallCodesphereCmd) RunE(_ *cobra.Command, _ []string) error {
 	}
```
OliverTrautvetter, line 73: Why do we use c.Opts here and everything else uses effectiveOpts? So use effectiveOpts here too for consistency?

OliverTrautvetter (approved):
approved 🚀 just added 2 small comments

cli/cmd/install_codesphere.go
nathanael-ruf: Unrelated to this PR though? This line was not touched.

cli/cmd/install_codesphere.go
nathanael-ruf: Resolving since unrelated. Maybe Tim can follow up.

cli/cmd/install_codesphere.go
nathanael-ruf: But @schrodit for second look.

OliverTrautvetter: Yes unrelated

PR #529, opened 2026-06-24
feat(gcp-bootstrap): Create an IP and configure SSH to use it for LB svc

nathanael-ruf:
CU-869dv8e9n

See https://codesphere-cloud.slack.com/archives/C0B4V6UF2QH/p1782288154121389?thread_ts=1782285831.692179&cid=C0B4V6UF2QH for my general approach.

Tried gcp bootstrap locally and the IP is created and the config looks correct to me.
The [corresponding change in the ssh chart](https://github.com/codesphere-cloud/charts/pull/747) is not merged yet, so can't try end to end, but worst case I'll create a followup PR.

Screenshot 2026-06-24 at 10 35 35
Screenshot 2026-06-24 at 10 36 18


schrodit (commented):
this PR only creates a new ip right?
The pc-apps chart is not yet configured, not sure if this PR makes sense without the pc-config update.
I think here its important how the data gets propagated to the chart

NautiluX (approved):

PR #531, opened 2026-06-25
feat: Add command to load images into custom registry

NautiluX:


cli/cmd/install_k0s_test.go
```diff
@@ -528,6 +528,7 @@ var _ = Describe("InstallK0sCmd", func() {
 				c.FileWriter = mockFileWriter
 
 				mockEnv.EXPECT().GetOmsWorkdir().Return(tempDir)
+				mockFileWriter.EXPECT().MkdirAll(tempDir, os.FileMode(0755)).Return(nil)
```
NautiluX, line 531: Unrelated fix that was necessary to run the tests locally.Seems they aren't executed in the CI

PR #539, opened 2026-06-29
fix(vault): trim trailing newlines from kubeconfig before storing in vault add unwrapSOPSData function to handle SOPS data extraction

OliverTrautvetter:
This pull request introduces several improvements and fixes related to handling kubeconfig storage, SOPS-encrypted vault data, and associated tests. The main changes ensure that kubeconfig files are stored without unwanted trailing newlines, add robust handling for SOPS-encrypted YAML files by unwrapping top-level `data` wrappers, and enhance testing to validate these behaviors.

**Kubeconfig storage improvements:**

* Trailing newlines are now trimmed from the kubeconfig content before storing it in the vault, preventing YAML formatting issues such as the use of `|+` chomping.
* Tests were updated and extended to verify that trailing newlines are properly removed and to ensure the vault file does not contain unwanted YAML formatting.

**SOPS/YAML handling improvements:**

* Added the `unwrapSOPSData` function to automatically strip a top-level `data` block scalar wrapper from SOPS-encrypted YAML files, ensuring the vault parser receives the intended document structure.
* Integrated the unwrapping logic into the vault data parser and made the function available for testing.
* Added comprehensive unit tests for `unwrapSOPSData` to verify correct behavior with various YAML inputs, including normal, wrapped, and invalid documents.

[Clickup](https://app.clickup.com/t/24560134/869dwweda)


internal/installer/export_test.go
```diff
@@ -64,3 +64,5 @@ func (o *OpenBaoInstaller) ValidateConfig() error {
 func (o *OpenBaoInstaller) ReadinessTimeout() time.Duration {
 	return o.readinessTimeout()
 }
+
+var UnwrapSOPSData = unwrapSOPSData
```
joka134: That's actually the same as simply exporting it. Can we avoid this? The related test scenario should be part of a new test file for internal tests. Just like we did in other packages as well.

internal/installer/vault_encryption.go
```diff
@@ -182,3 +183,27 @@ func DecryptFileWithSOPS(src, keyPath string) ([]byte, error) {
 	}
 	return out, nil
 }
+
+// unwrapSOPSData strips a top-level "data" literal block scalar wrapper if
+// present. When SOPS encrypts with --input-type yaml, it
```
joka134, line 188: Why didn't we see this issue before when using sops? 🤔

joka134 (changes requested):

internal/installer/vault_encryption.go
OliverTrautvetter: Before we did not use --input-type, which is the correct way to handle YAML files. It should have used the data block also before recent changes. Yet --decrypt should remove that data thing, but for some reasons that can fail and now unwrapSOPSData is the safety function to fix it if --decrypt does not remove it for some reasons

internal/installer/export_test.go
OliverTrautvetter: moved to another file

internal/installer/vault_encryption_unexported_test.go
joka134, line 1: name should be `vault_encryption_unexported_test.go`. Aligns with what we have already

joka134 (changes requested):
Looks good. One simple naming comment

joka134 (approved):

PR #544, opened 2026-06-30
refac(cs-flags): Split experiments into internal and preview

nathanael-ruf:
Also enabled `workspace-ssh` preview flag by default (was handed over to SRE yesterday).


cli/cmd/bootstrap_gcp.go
```diff
@@ -28,7 +28,9 @@ type BootstrapGcpCmd struct {
 	CodesphereEnv     *gcp.CodesphereEnvironment
 	InputRegistryType string
 	SSHQuiet          bool
-	FeatureFlagList   []string
+	// Experiments backs the deprecated --experiments flag; its values
+	// are folded into the internal bucket for backwards compatibility.
+	Experiments []string
```
nathanael-ruf: Not sure if/who uses that flag. Up to you to eventually remove it.

internal/bootstrap/gcp/gcp.go
```diff
@@ -87,41 +97,42 @@ type GCPBootstrapper struct {
...
+	SshProxyIP                    string       `json:"ssh_proxy_ip"`
+	RegistryType                  RegistryType `json:"registry_type"`
+	GitHubPAT                     string       `json:"-"`
+	GitHubAppName                 string       `json:"-"`
+	GitHubTeamOrg                 string       `json:"github_team_org"`
+	GitHubTeamSlug                string       `json:"github_team_slug"`
+	RegistryUser                  string       `json:"-"`
+	InternalFlags                 []string     `json:"internal"`
+	PreviewFlags                  []string     `json:"preview"`
+	FeatureFlags                  []string     `json:"feature_flags"`
```
nathanael-ruf, line 130: Decided to make them all slices internally. I think that is easier to understand and it's good enough to convert to a map where needed when writing the config file.
Not sure why it's even inconsistent in the config.yml file.

internal/bootstrap/gcp/gcp.go
```diff
@@ -62,15 +62,25 @@ func GetDNSRecordNames(baseDomain string) []struct {
...
+// development and not intended for customer use.
+// Atm. it's not empty as the internal flags below are likely preview or
+// feature flags, but are still in the internal bucket for historical
+// reasons (before we only had one "experiments" bucket).
+var DefaultInternalFlags []string = []string{
 	"headless-services",
 	"vcluster",
 	"custom-service-image",
 	"ms-in-ls",
+}
```
OliverTrautvetter, line 75: "This should ALWAYS be empty" is kind of misleading, because it looks like it can never be empty because of previous problems. Can we move these flags to their correct buckets. Something like DefaultPreviewFlags or DefaultFeatureFlags

cli/cmd/bootstrap_gcp.go
```diff
@@ -172,8 +177,8 @@ func (c *BootstrapGcpCmd) BootstrapGcp() error {
 		}
 	}
 
-	for _, flag := range c.FeatureFlagList {
-		c.CodesphereEnv.FeatureFlags[flag] = true
+	if c.cmd.Flags().Changed("experiments") && !c.cmd.Flags().Changed("internal-flags") {
+		c.CodesphereEnv.InternalFlags = c.Experiments
 	}
```
OliverTrautvetter: Silently drops fields if both -experiments foo and --internal-flags are set. We should log what is removed.

and maybe log a warning if both of the flags are set at the same time.

cli/cmd/bootstrap_local.go
```diff
@@ -124,8 +129,8 @@ func (c *BootstrapLocalCmd) BootstrapLocal() error {
 		return err
 	}
 
-	for _, flag := range c.FeatureFlagList {
-		c.CodesphereEnv.FeatureFlags[flag] = true
+	if c.cmd.Flags().Changed("experiments") && !c.cmd.Flags().Changed("internal-flags") {
+		c.CodesphereEnv.InternalFlags = c.Experiments
```
OliverTrautvetter: same comment as cli/cmd/bootstrap_gcp.go

internal/installer/files/config_yaml.go
```diff
@@ -302,7 +302,8 @@ type CodesphereConfig struct {
 	CertIssuer                 CertIssuerConfig       `yaml:"certIssuer"`
 	CustomDomains              CustomDomainsConfig    `yaml:"customDomains"`
 	DNSServers                 []string               `yaml:"dnsServers"`
-	Experiments                []string               `yaml:"experiments"`
+	Internal                   []string               `yaml:"internal"`
+	Preview                    map[string]bool        `yaml:"preview"`
```
OliverTrautvetter, line 306: can we do that consistent? So internal also a map[string]bool? Otherwise the YAML format depends on which bucket a flag lives in. That might be confusing

cli/cmd/bootstrap_gcp.go
```diff
@@ -28,7 +28,9 @@ type BootstrapGcpCmd struct {
 	CodesphereEnv     *gcp.CodesphereEnvironment
```
OliverTrautvetter, line 28: description says "Also enabled workspace-ssh preview flag by default (was handed over to SRE yesterday)."

I don't see any changes to workspace-ssh? What do you mean by that?

cli/cmd/bootstrap_local.go
```diff
@@ -35,10 +35,12 @@ import (
...
-	cmd             *cobra.Command
-	CodesphereEnv   *local.CodesphereEnvironment
-	Yes             bool
-	FeatureFlagList []string
+	cmd           *cobra.Command
+	CodesphereEnv *local.CodesphereEnvironment
+	Yes           bool
+	// Experiments backs the deprecated --experiments flag; its values
+	// are folded into the internal bucket for backwards compatibility.
+	Experiments []string
```
OliverTrautvetter: can be unexported, so experiments. It is not used outside the package

internal/util/map.go
```diff
@@ -3,6 +3,20 @@
 
 package util
 
+// StringSliceToBoolMap converts a slice of strings into a set-style map where
+// every element maps to true. A nil slice yields a nil map; a non-nil (possibly
+// empty) slice yields a non-nil map.
+func StringSliceToBoolMap(items []string) map[string]bool {
```
OliverTrautvetter, line 9: very nice helper, looks good

OliverTrautvetter (changes requested):
Looks very nice in general 👍 added some comments

internal/bootstrap/gcp/gcp.go
nathanael-ruf: We can't, because they are actually still internal flags in Codesphere itself. So this is an unrelated change (which we still should do, but I don't think it should block this PR).

internal/installer/files/config_yaml.go
nathanael-ruf: I agree that this is confusing, but it's not up to OMS here. The config.yaml format is dictated by the js pc installer. I already tried to use the same data structure in OMS where possible (see [this](https://github.com/codesphere-cloud/oms/pull/544#discussion_r3497073566)), but this is the last place where we can (and must) convert.

cli/cmd/bootstrap_gcp.go
nathanael-ruf: Ups, forgot to add after all.
Basically me wanting to add workspace-ssh as default triggered this whole PR, but then I forgot to add it. Done now.

internal/bootstrap/gcp/gcp.go
OliverTrautvetter: Ok, as long as we keep that in mind 👍

OliverTrautvetter (approved):
🚀

PR #545, opened 2026-06-30
feat(openbao): add imagePullSecrets + use mirrored bank vaults operator

Jcing95:
## What

Makes the `oms install openbao` image and chart sources configurable for mirrored OCI registries, and adds an image pull secret so the OpenBao, bank-vaults, and operator images can be pulled from a private registry.

## Why

The OpenBao server, bank-vaults configurer, and operator images already ship from the Codesphere GHCR mirror, but there was no way to authenticate pulls against a private registry, and the operator Helm chart was still pulled from public `ghcr.io/bank-vaults`. This lets a cluster (or a customer using their own mirror) pull every artifact from a single private registry.

## Changes

- **New CLI flags** (with defaults pointing at the Codesphere mirror): `--openbao-image`, `--bank-vaults-image`, `--operator-image`, `--operator-chart-repo`. Empty values are backfilled from the `Default*` constants in `validateConfig`, so programmatic callers and the default install are unaffected.
- **Image pull secret**: when `OMS_REGISTRY_USER` and `OMS_REGISTRY_PASSWORD` are both set, the installer creates/updates a `dockerconfigjson` secret (`openbao-registry`) with one `auths` entry per distinct registry host derived from the configured image refs. It is:
  - attached to the `openbao` ServiceAccount (via the Vault CR template), and
  - passed to the operator pod through the chart's `image.imagePullSecrets` values.
  - Setting only one of the two credentials is a hard error; setting neither is a no-op (unchanged behavior).
  - Creation is idempotent and refreshes credentials on re-run, with create/update conflict handling.
- **Helm OCI login**: before installing/upgrading the operator chart, the installer authenticates to the chart registry host (when credentials are present), mirroring the `pc_apps.go` pattern, so the chart can be pulled from a private OCI repo.
- **Operator chart/image bump**: vault-operator chart `1.22.5` → `1.24.0`, default operator image `vault-operator:1.24.0`. Default operator chart repo now points at the Codesphere mirror (`oci://ghcr.io/codesphere-cloud/docker/ghcr.io/bank-vaults/helm-charts`).
- **Docs**: regenerated `docs/oms_install_openbao.md` with the new flags, the env vars, and a mirrored-registry example.

## Notes for reviewers

- Because the default operator chart repo now points at the mirror, the Helm chart pull (client-side) requires `OMS_REGISTRY_USER`/`OMS_REGISTRY_PASSWORD` unless the mirror packages are public-read. Node-level registry creds do not help the client-side chart pull.
- The operator chart values use `image.repository` / `image.tag` / `image.imagePullSecrets` per the vault-operator chart schema; a chart version bump could move these keys (wrong keys are silently ignored and fall back to chart defaults). Verify with `helm show values oci://ghcr.io/bank-vaults/helm-charts/vault-operator --version 1.24.0`.

## Testing

- `go build ./...`, `go test ./internal/installer/...` pass; `gofmt` clean.
- New tests cover: mirror overrides + pull secret wiring, secret creation/no-op/partial-cred error/idempotency, host deduplication, `validateConfig` default backfill, and ServiceAccount `imagePullSecrets` template rendering (set and unset).

### DIFF
--- a/cli/cmd/install_openbao.go
+++ b/cli/cmd/install_openbao.go
@@ -79,10 +79,10 @@ func (c *InstallOpenBaoCmd) RunE(_ *cobra.Command, _ []string) error {
 		Timeout:           c.Opts.Timeout,
 		AgeRecipient:      recipient,
 		AgeKeyPath:        keyPath,
-		// Optional GHCR credentials for the private OpenBao/bank-vaults image
-		// mirror. When both are set the installer creates a pull secret and
-		// wires it onto the openbao ServiceAccount; when unset, behavior is
-		// unchanged.
+		// Optional registry credentials for the private OpenBao/bank-vaults image
+		// mirror (any OCI registry, not just ghcr.io). When both are set the
+		// installer creates a pull secret and wires it onto the openbao
+		// ServiceAccount; when unset, behavior is unchanged.
 		RegistryUser:     os.Getenv("OMS_REGISTRY_USER"),
 		RegistryPassword: os.Getenv("OMS_REGISTRY_PASSWORD"),
 		// Image/chart overrides for mirrored OCI registries. Defaults are set on
--- a/cli/cmd/install_openbao.go
+++ b/cli/cmd/install_openbao.go
@@ -167,7 +167,7 @@ func AddInstallOpenBaoCmd(install *cobra.Command, opts *GlobalOptions) {
 				{Cmd: "--dr-backup-path ./backups/cluster-1.enc.json", Desc: "Fresh bootstrap with DR backup saved locally"},
 				{Cmd: "--dr-backup-path ./backups/cluster-1.enc.json --secrets-engine my-engine --bao-user myuser", Desc: "Custom engine and user"},
 				{Cmd: "--dr-backup-path ./backups/cluster-1.enc.json --timeout 10m", Desc: "Extended timeout for slower clusters"},
-				{Cmd: "--dr-backup-path ./backups/cluster-1.enc.json --openbao-image my-mirror.example.com/openbao/openbao:2.5.4 --operator-chart-repo oci://my-mirror.example.com/bank-vaults/helm-charts", Desc: "Use a mirrored OCI registry (set OMS_REGISTRY_USER/OMS_REGISTRY_PASSWORD)"},
+				{Cmd: "--dr-backup-path ./backups/cluster-1.enc.json --openbao-image my-mirror.example.com/openbao/openbao:2.5.5 --operator-chart-repo oci://my-mirror.example.com/bank-vaults/helm-charts", Desc: "Use a mirrored OCI registry (set OMS_REGISTRY_USER/OMS_REGISTRY_PASSWORD)"},
 			}),
 		},
 		Opts: &InstallOpenBaoOpts{GlobalOptions: opts},
--- a/cli/cmd/install_openbao.go
+++ b/cli/cmd/install_openbao.go
@@ -44,6 +44,10 @@ type InstallOpenBaoOpts struct {
 	Timeout           time.Duration
 	AgeKeyFile        string
 	Yes               bool
+	OpenBaoImage      string
+	BankVaultsImage   string
+	OperatorImage     string
+	OperatorChartRepo string
 }
 
 func (c *InstallOpenBaoCmd) RunE(_ *cobra.Command, _ []string) error {
@@ -75,6 +79,18 @@ func (c *InstallOpenBaoCmd) RunE(_ *cobra.Command, _ []string) error {
 		Timeout:           c.Opts.Timeout,
 		AgeRecipient:      recipient,
 		AgeKeyPath:        keyPath,
+		// Optional GHCR credentials for the private OpenBao/bank-vaults image
+		// mirror. When both are set the installer creates a pull secret and
+		// wires it onto the openbao ServiceAccount; when unset, behavior is
+		// unchanged.
+		RegistryUser:     os.Getenv("OMS_REGISTRY_USER"),
+		RegistryPassword: os.Getenv("OMS_REGISTRY_PASSWORD"),
+		// Image/chart overrides for mirrored OCI registries. Defaults are set on
+		// the flags (the installer's Default* values).
+		OpenBaoImage:      c.Opts.OpenBaoImage,
+		BankVaultsImage:   c.Opts.BankVaultsImage,
+		OperatorImage:     c.Opts.OperatorImage,
+		OperatorChartRepo: c.Opts.OperatorChartRepo,
 	}
 
 	inst, err := installer.NewOpenBaoInstaller(cfg)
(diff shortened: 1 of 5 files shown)


internal/installer/openbao.go
```diff
@@ -590,6 +721,125 @@ func (o *OpenBaoInstaller) ensureUnsealSecret(secretsClient corev1client.SecretI
...
+// every registry the install pulls from. It is attached to the openbao
+// ServiceAccount by ApplyVaultCR and to the operator chart by
+// DeployBankVaultsOperator when credentials are present.
+func (o *OpenBaoInstaller) EnsureImagePullSecret() error {
+	if o.Config.RegistryUser == "" && o.Config.RegistryPassword == "" {
+		return nil
+	}
+	if !o.imagePullSecretConfigured() {
+		return fmt.Errorf("incomplete registry credentials: set both OMS_REGISTRY_USER and OMS_REGISTRY_PASSWORD, or neither")
+	}
```
OliverTrautvetter: Can be simplified.

imagePullSecretConfigured already checks `o.Config.RegistryUser != "" && o.Config.RegistryPassword != ""`

internal/installer/openbao.go
```diff
@@ -218,6 +262,13 @@ func (o *OpenBaoInstaller) Install(ctx context.Context) error {
 		return fmt.Errorf("failed to ensure namespace: %w", err)
 	}
 
+	// Create the GHCR pull secret before the operator schedules any pods, so
```
OliverTrautvetter: is it working just for GHCR? Does it not work for any registry host derived from the image refs?

cli/cmd/install_openbao.go
```diff
@@ -146,6 +180,10 @@ func AddInstallOpenBaoCmd(install *cobra.Command, opts *GlobalOptions) {
...
 	openbao.cmd.Flags().StringVarP(&openbao.Opts.AgeKeyFile, "age-key-file", "k", "", "Path to age private key file for SOPS encryption/decryption (auto-detected if not set)")
 	openbao.cmd.Flags().BoolVarP(&openbao.Opts.Yes, "yes", "y", false, "Auto-approve re-initialization of an existing deployment when no DR backup is found")
+	openbao.cmd.Flags().StringVar(&openbao.Opts.OpenBaoImage, "openbao-image", installer.DefaultOpenBaoImage, "OpenBao server image (override for a mirrored OCI registry)")
+	openbao.cmd.Flags().StringVar(&openbao.Opts.BankVaultsImage, "bank-vaults-image", installer.DefaultBankVaultsImage, "Bank-Vaults configurer image (override for a mirrored OCI registry)")
+	openbao.cmd.Flags().StringVar(&openbao.Opts.OperatorImage, "operator-image", installer.DefaultOperatorImage, "Bank-Vaults operator pod image (override for a mirrored OCI registry)")
+	openbao.cmd.Flags().StringVar(&openbao.Opts.OperatorChartRepo, "operator-chart-repo", installer.DefaultBankVaultsChartRepo, "OCI repo hosting the vault-operator Helm chart (override for a mirrored OCI registry)")
```
OliverTrautvetter, line 187: Should we check if it starts with oci:// ? In validateConfig()

something like 
```
if !strings.HasPrefix(o.Config.OperatorChartRepo, "oci://") {
    return fmt.Errorf("--operator-chart-repo must use oci:// scheme, got %q", o.Config.OperatorChartRepo)
}
```

internal/installer/openbao.go
```diff
@@ -365,9 +423,70 @@ func (o *OpenBaoInstaller) DeployBankVaultsOperator() error {
...
+	image := map[string]interface{}{}
+	if o.Config.OperatorImage != "" {
+		ref, err := reference.ParseNormalizedNamed(o.Config.OperatorImage)
+		if err != nil {
+			return nil, fmt.Errorf("parsing operator image %q: %w", o.Config.OperatorImage, err)
+		}
+		image["repository"] = reference.TrimNamed(ref).Name()
+		if tagged, ok := ref.(reference.Tagged); ok {
+			image["tag"] = tagged.Tag()
+		}
```
OliverTrautvetter: Maybe check if a tag is set. So return an error if the image reference has neither a tag nor a digest

internal/installer/openbao.go
```diff
@@ -590,6 +721,125 @@ func (o *OpenBaoInstaller) ensureUnsealSecret(secretsClient corev1client.SecretI
...
+		auths[host] = dockerAuth{Username: username, Password: password, Auth: auth}
+	}
+	return json.Marshal(map[string]map[string]dockerAuth{"auths": auths})
+}
+
+// registryHostsFor returns the distinct registry hosts of the given image
+// references (e.g. "ghcr.io/codesphere-cloud/.../openbao:2.5.4" -> "ghcr.io").
+// Empty refs are skipped. The result is order-stable so the rendered secret is
+// deterministic across runs.
+func registryHostsFor(images ...string) ([]string, error) {
```
OliverTrautvetter, line 832: Maybe add some test for digest-only references (like registry.example.com/image@sha256:abc...)

OliverTrautvetter (changes requested):
Looks mostly good 👍 some smaller comments

cli/cmd/install_openbao.go
```diff
@@ -75,6 +79,18 @@ func (c *InstallOpenBaoCmd) RunE(_ *cobra.Command, _ []string) error {
 		Timeout:           c.Opts.Timeout,
 		AgeRecipient:      recipient,
 		AgeKeyPath:        keyPath,
+		// Optional GHCR credentials for the private OpenBao/bank-vaults image
+		// mirror. When both are set the installer creates a pull secret and
+		// wires it onto the openbao ServiceAccount; when unset, behavior is
+		// unchanged.
```
OliverTrautvetter: Nit: the comment still says "GHCR credentials"

It should say "Registry credentials" since the feature works with any OCI registry, not just ghcr.io.

internal/installer/openbao.go
```diff
@@ -325,15 +384,19 @@ func (o *OpenBaoInstaller) GeneratePassword() error {
 // installed in a different namespace, we skip re-deployment — one instance
 // is sufficient for the entire cluster.
 func (o *OpenBaoInstaller) DeployBankVaultsOperator() error {
```
OliverTrautvetter, line 389: We should add a test case that verifies LoginRegistry is called on the Helm mock when RegistryUser/RegistryPassword are set. Right now no test exercises the authenticated path.

internal/installer/openbao.go
```diff
@@ -366,9 +432,72 @@ func (o *OpenBaoInstaller) DeployBankVaultsOperator() error {
...
+	image := map[string]interface{}{}
+	if o.Config.OperatorImage != "" {
+		ref, err := reference.ParseNormalizedNamed(o.Config.OperatorImage)
+		if err != nil {
+			return nil, fmt.Errorf("parsing operator image %q: %w", o.Config.OperatorImage, err)
+		}
+		image["repository"] = reference.TrimNamed(ref).Name()
+		tagged, ok := ref.(reference.Tagged)
+		if !ok {
+			return nil, fmt.Errorf("operator image %q must include a tag: the vault-operator chart renders the image as repository:tag, so digest-only references are not supported", o.Config.OperatorImage)
```
OliverTrautvetter, line 489: Nice error, please add that explanation also to the doc in OpenBaoInstallerConfig

OliverTrautvetter (changes requested):
Looks very good, just some minor comments now, but almost done 👍

PR #553, opened 2026-07-01
feat(fga): OMS command to initialize first cluster admin

nbrodnicke:
CU-869dy61ny

Adds a new oms add-cluster-admin CLI command for bootstrapping the first cluster admin on a Codesphere installation.

The command writes the given email address into a Kubernetes secret (cluster-admin-email by default) in the target cluster. On startup, the AuthService reads this secret and grants the stored email cluster-admin permissions via OpenFGA — this is how a fresh installation gets its first admin without a chicken-and-egg problem (no admin exists yet to grant one via the normal permission flow).

Behavior:

Creates the secret if it doesn't exist, or updates it if it does (idempotent — re-running with the same email is a no-op).
--email (required): the cluster admin's email address, validated per RFC 5322 and normalized (lowercased). Malformed input, e.g. not-an-email or a display name without angle brackets like Max Mustermann max@mail.com, is rejected up front with a clear error rather than propagating downstream.
--namespace (default codesphere) and --secret-name (default cluster-admin-email): overridable in case of custom deployments. Empty/whitespace values are rejected early instead of silently overwriting the sensible default.
The target cluster is the current kubeconfig context; can be pointed elsewhere via the standard KUBECONFIG environment variable.

### DIFF
--- a/cli/cmd/add_cluster_admin.go
+++ b/cli/cmd/add_cluster_admin.go
@@ -41,7 +41,8 @@ func AddAddClusterAdminCmd(parent *cobra.Command, opts *GlobalOptions) {
 				deployment consumes via a secretKeyRef. The secret is created if it does not exist yet and
 				updated otherwise, so running the command again overwrites the previous email.
 
-				The target cluster is determined by the current kubeconfig context.`),
+				The target cluster is determined by the current kubeconfig context. Set the KUBECONFIG
+				environment variable to target a different kubeconfig.`),
 			Example: formatExamples("add-cluster-admin", []packageio.Example{
 				{Cmd: "--email niklas@codesphere.com", Desc: "Set the cluster admin email using the default secret and namespace"},
 				{Cmd: "--email admin@codesphere.com --namespace kube-system --secret-name cluster-admin-email", Desc: "Set the cluster admin email in a custom namespace"},
--- a/cli/cmd/add_cluster_admin.go
+++ b/cli/cmd/add_cluster_admin.go
@@ -1,0 +1,62 @@
+// Copyright (c) Codesphere Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"fmt"
+
+	packageio "github.com/codesphere-cloud/cs-go/pkg/io"
+	"github.com/codesphere-cloud/oms/internal/clusteradmin"
+	"github.com/codesphere-cloud/oms/internal/util"
+	"github.com/spf13/cobra"
+)
+
+type AddClusterAdminCmd struct {
+	cmd  *cobra.Command
+	Opts AddClusterAdminOpts
+}
+
+type AddClusterAdminOpts struct {
+	*GlobalOptions
+	clusteradmin.Opts
+}
+
+func (c *AddClusterAdminCmd) RunE(_ *cobra.Command, _ []string) error {
+	clientset, _, err := util.NewClients()
+	if err != nil {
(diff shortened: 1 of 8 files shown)


cli/cmd/add_cluster_admin.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Codesphere Inc.
```
OliverTrautvetter, line 1: The PR body is just CU-869dy61ny. Please add a short summary explaining the context and what problem this solves.

internal/clusteradmin/clusteradmin.go
```diff
@@ -0,0 +1,99 @@
...
+
+	if _, err := secrets.Update(ctx, existing, metav1.UpdateOptions{}); err != nil {
+		return fmt.Errorf("updating secret %s/%s: %w", opts.Namespace, opts.SecretName, err)
+	}
+	log.Printf("Set cluster admin email '%s' in secret '%s/%s'", email, opts.Namespace, opts.SecretName)
+	return nil
+}
+
+// normalizeEmail validates and canonicalizes an email address.
+func normalizeEmail(raw string) (string, error) {
```
OliverTrautvetter, line 96: maybe add some test for a mail input like 
 --email "Max Mustermann <max@mail.com>"

cli/cmd/add_cluster_admin.go
```diff
@@ -0,0 +1,62 @@
...
+			}),
+		},
+		Opts: AddClusterAdminOpts{GlobalOptions: opts},
+	}
+	c.cmd.RunE = c.RunE
+
+	flags := c.cmd.Flags()
+	flags.StringVar(&c.Opts.Email, "email", "", "Email address of the cluster admin (required)")
+	flags.StringVar(&c.Opts.Namespace, "namespace", clusteradmin.DefaultNamespace, "Kubernetes namespace where the secret is stored")
+	flags.StringVar(&c.Opts.SecretName, "secret-name", clusteradmin.DefaultSecretName, "Name of the Kubernetes secret holding the cluster admin email")
```
OliverTrautvetter, line 58: We should validate the input for namespace and secret-name

they have defaults but a user could still enter this
--namespace ""
--secret-name ""

cli/cmd/add_cluster_admin.go
```diff
@@ -0,0 +1,62 @@
...
+	c := AddClusterAdminCmd{
+		cmd: &cobra.Command{
+			Use:   "add-cluster-admin",
+			Short: "Set the cluster admin email in a Kubernetes secret",
+			Long: packageio.Long(`Sets the cluster admin email in the target Kubernetes cluster by writing it to a
+				Kubernetes secret. The email is stored under the 'email' key of the secret, which the platform
+				deployment consumes via a secretKeyRef. The secret is created if it does not exist yet and
+				updated otherwise, so running the command again overwrites the previous email.
+
+				The target cluster is determined by the current kubeconfig context.`),
```
OliverTrautvetter: should we also add an optional flag --kubeconfig?

OliverTrautvetter (changes requested):
Looks good, just some small comments 👍

cli/cmd/add_cluster_admin.go
nbrodnicke: I think you can set the kubeconfig via a global variable, and think that should be enough here, but thanks for the idea :)

OliverTrautvetter (approved):
🚀

PR #555, opened 2026-07-03
feat(fga): Cluster Admin Email assignment gets part of bootstrap command

nbrodnicke:
CU-869dz4umw

Cluster admin email as part of bootstrap-gcp

Previously, the cluster admin email was set manually in a separate SSH step after bootstrap. This PR makes it a regular, optional part of the bootstrap and install flow instead.

What changes:

New --cluster-admin-email flag on oms beta bootstrap-gcp (optional, no behavior change when unset).
The email is now stored in the install config (codesphere.clusterAdminEmail) instead of only being set temporarily via SSH — it therefore survives re-runs with --write-config/--recover-config.
oms install codesphere automatically creates the cluster-admin-email secret before the platform phase (auth-service) starts — no manual extra step, no service restart needed.
Reuses the existing add-cluster-admin logic instead of duplicating it.
Why: So integration tests (and production bootstraps) get the cluster admin set consistently and reproducibly from the start, instead of relying on a fragile, separate SSH step in the CI workflow.

### DIFF
--- a/cli/cmd/install_codesphere_platform.go
+++ b/cli/cmd/install_codesphere_platform.go
@@ -9,10 +9,14 @@ import (
 	"runtime"
 
 	"github.com/codesphere-cloud/cs-go/pkg/io"
+	"github.com/codesphere-cloud/oms/internal/clusteradmin"
 	"github.com/codesphere-cloud/oms/internal/env"
 	"github.com/codesphere-cloud/oms/internal/installer"
+	"github.com/codesphere-cloud/oms/internal/installer/files"
 	"github.com/codesphere-cloud/oms/internal/system"
+	"github.com/codesphere-cloud/oms/internal/util"
 	"github.com/spf13/cobra"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 // InstallCodespherePlatformCmd runs only the Codesphere platform step (Phase 3).
@@ -23,16 +27,20 @@ type InstallCodespherePlatformCmd struct {
 }
 
 func (c *InstallCodespherePlatformCmd) RunE(_ *cobra.Command, _ []string) error {
-	effectiveOpts, _, cleanup, err := prepareInstallConfig(c.Opts, installer.NewConfig())
+	effectiveOpts, cfg, cleanup, err := prepareInstallConfig(c.Opts, installer.NewConfig())
 	if err != nil {
 		return err
 	}
 	defer cleanup()
 
-	return installCodespherePlatform(effectiveOpts, c.Env)
+	return installCodespherePlatform(effectiveOpts, cfg, c.Env)
 }
 
-func installCodespherePlatform(opts *InstallCodesphereOpts, env env.Env) error {
+func installCodespherePlatform(opts *InstallCodesphereOpts, cfg files.RootConfig, env env.Env) error {
+	if err := ensureClusterAdminSecret(context.Background(), opts, cfg); err != nil {
+		return fmt.Errorf("failed to set cluster admin email: %w", err)
+	}
+
 	workdir := env.GetOmsWorkdir()
 	pm := installer.NewPackage(workdir, opts.Package)
 	cm := installer.NewConfig()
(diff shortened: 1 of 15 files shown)


cli/cmd/install_codesphere_platform.go
```diff
@@ -23,16 +27,20 @@ type InstallCodespherePlatformCmd struct {
...
 
-	return installCodespherePlatform(effectiveOpts, c.Env)
+	return installCodespherePlatform(effectiveOpts, cfg, c.Env)
 }
 
-func installCodespherePlatform(opts *InstallCodesphereOpts, env env.Env) error {
+func installCodespherePlatform(opts *InstallCodesphereOpts, cfg files.RootConfig, env env.Env) error {
+	if err := ensureClusterAdminSecret(context.Background(), opts, cfg); err != nil {
+		return fmt.Errorf("failed to set cluster admin email: %w", err)
+	}
```
OliverTrautvetter: Could we thread a context.Context through installCodespherePlatform instead of using a fixed context.Background()?

internal/bootstrap/gcp/gcp.go
```diff
@@ -438,9 +440,34 @@ func (b *GCPBootstrapper) ValidateInput() error {
...
+	}
+
+	email, err := clusteradmin.NormalizeEmail(b.Env.ClusterAdminEmail)
+	if err != nil {
+		return fmt.Errorf("invalid cluster admin email: %w", err)
+	}
+	b.Env.ClusterAdminEmail = email
+
+	return nil
+}
```
OliverTrautvetter, line 469: Should we allow the flag without --write-config when --recover-config is also set?

cli/cmd/install_codesphere_platform.go
```diff
@@ -55,6 +63,45 @@ func installCodespherePlatform(opts *InstallCodesphereOpts, env env.Env) error {
 	return nil
 }
 
+// ensureClusterAdminSecret applies the cluster admin email configured via
+// codesphere.clusterAdminEmail to the cluster-admin-email secret before the
+// platform is installed, so the auth-service finds it on first start.
+// It is a no-op when the config does not set an email.
+func ensureClusterAdminSecret(ctx context.Context, opts *InstallCodesphereOpts, cfg files.RootConfig) error {
```
OliverTrautvetter: This looks like it duplicates the vault-loading and kubeconfig-extraction pattern that already exists in loadVaultData in install_codesphere_dependencies.go

extract a shared helper maybe?

internal/clusteradmin/clusteradmin.go
```diff
@@ -92,8 +102,19 @@ func AddClusterAdmin(ctx context.Context, clientset kubernetes.Interface, opts O
...
+// ensureNamespace creates the namespace if it does not exist yet.
+func ensureNamespace(ctx context.Context, clientset kubernetes.Interface, namespace string) error {
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{Name: namespace},
+	}
+	if _, err := clientset.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{}); err != nil && !apierrors.IsAlreadyExists(err) {
+		return fmt.Errorf("creating namespace %s: %w", namespace, err)
+	}
+	return nil
+}
```
OliverTrautvetter, line 114: We don't wait for the namespace, so the namespace might not be visible yet when secrets.Create is called. Is that a problem?

cli/cmd/install_codesphere_platform.go
OliverTrautvetter: Also this function does a lot of infrastructure stuff. So maybe this function would make more sense in the installer or clusteradmin package and not the cli command layer

OliverTrautvetter (changes requested):
Some comments and questions 👍

internal/bootstrap/gcp/gcp.go
nbrodnicke: If we would allow --recover-config, it uses the old config, which might not have the required cluster admin variable set. This variable gets into the config only via the UpdateInstallConfig step.

internal/clusteradmin/clusteradmin.go
nbrodnicke: CoreV1, Namespaces and Create are all synchronous functions, so we should need to wait for the namespace here. When the function is done, it should be there.

internal/bootstrap/gcp/gcp.go
OliverTrautvetter: You are right, that would be super annoying to handle

internal/installer/cluster_admin.go
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) Codesphere Inc.
```
OliverTrautvetter, line 1: The update looks very good, but can you please add some unit tests for this new shared internal functions / this new file?

OliverTrautvetter (changes requested):
Just 1 thing:

OliverTrautvetter (approved):
🚀 looks good

PR #578, opened 2026-07-14
fix: don't render configs twice

jonas-zipprick:
Instead, render them only once.
Rendering twice is bad, because double curly brackets like "{{" need to be escaped _twice_ if you want them to end up in the plain text.

The current flow in main is:

1. prepareInstallConfig(...) renders each input config with RenderConfigFileToTemp(...).
2. That merged temp file is then passed into ci.Install(...).
3. ci.Install(...) calls prepareConfig(...), which renders the config again if VaultPath != "".

This PR removes the second render run in the function prepareConfig.

### DIFF
--- a/internal/installer/codesphere.go
+++ b/internal/installer/codesphere.go
@@ -15,9 +15,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/codesphere-cloud/oms/internal/configtemplating"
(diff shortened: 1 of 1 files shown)


schrodit (approved):
looks good since we anyway have the prepareInstallConfig before this function is called

PR #580, opened 2026-07-16
chore(docs): Add installation documentation

NautiluX:


docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,396 @@
+# Prepare infrastructure and install Codesphere
+
+This guide describes how to prepare infrastructure for Codesphere without depending on a specific cloud provider.
+
+> The machine sizes below are minimums. A Codesphere installation should use **at least four Ceph nodes**.
+
+## What the installation builds
+
+The installation uses a management host as the single entry point into a private network. From there, OMS installs and configures PostgreSQL, Ceph, k0s, and the Codesphere platform.
```
joka134:
```suggestion
The installation uses a jumphost as the single entry point into a private network. From there, OMS installs and configures PostgreSQL, Ceph, k0s, and the Codesphere platform.
```
another name for jumpbox/jumphost/controller? 😄

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,413 @@
...
+export OMS_PORTAL_API_KEY='<portal-api-key>'
+```
+
+The API key must remain set for the later package-listing and download commands. Supply it through an approved secret manager or protected session setup, and do not commit it to a shell profile, image, or source repository.
+
+## 8. Prepare access to GitHub Container Registry
+
+Codesphere images are pulled directly from GitHub Container Registry. Every k0s node must have outbound HTTPS access to `ghcr.io` and the related GitHub package endpoints.
+
+Create or select a GitHub personal access token with at least the `read:packages` scope and access to the Codesphere packages. If the organization enforces SSO, authorize the token for that organization. Use a dedicated machine account where possible so image pulls do not depend on an employee account.
```
clemensfreitagdev: in a general case OMS user cannot create those, so we such rather describe that you get this from us?

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,413 @@
...
+oms init install-config \
+  --profile production \
+  --config config.yaml \
+  --vault prod.vault.yaml \
+  --with-comments
+```
+
+The profile supplies initial values and the wizard prompts for the final settings. Use `production` for the topology in this guide. The `minimal` profile is intended only for constrained test installations and does not reduce the minimum hardware requirements stated above.
+
+An Ansible inventory can prepopulate the Kubernetes and Ceph host lists. For example:
```
clemensfreitagdev: should describe what happens if you would not have such a file

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,413 @@
...
+  ```
+
+  For a highly available control plane, list at least three addresses under `controlPlanes`. Dedicated worker addresses go under `workers`; size that list for the expected workload CPU, memory, and failure tolerance.
+
+  The interactive `oms init install-config` wizard asks separately for the comma-separated control-plane and worker IPs. With an Ansible inventory, put controller hosts under `k8s-cp` and worker hosts under `k8s-workers`.
+
+  k0s supports combined control-plane/worker nodes. However, the current OMS k0sctl configuration generator treats an IP present in both lists as control-plane-only and ignores the duplicate worker entry. Until combined-role generation is supported by OMS, use dedicated control-plane and worker entries in this workflow.
+- The platform and public gateways use `LoadBalancer`, or `ExternalIP` where that is the chosen integration.
+- `codesphere.domain` is `cs.<base-domain>`.
+- `codesphere.workspaceHostingBaseDomain` and the custom-domain CNAME base are `ws.<base-domain>`.
```
clemensfreitagdev: for the workspaceBaseDomain the de-facto standard we use is <dcId>.<baseDomain>, I'd put this in the Guide as well. The dcId shall be part of it.

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,439 @@
...
+export OMS_PORTAL_API_KEY='<portal-api-key>'
+```
+
+The API key must remain set for subsequent OMS portal operations. Supply it through an approved secret manager or protected session setup, and do not commit it to a shell profile, image, or source repository.
+
+## 8. Prepare access to GitHub Container Registry
+
+Codesphere images are pulled directly from GitHub Container Registry. Every k0s node must have outbound HTTPS access to `ghcr.io` and the related GitHub package endpoints.
+
+Create or select a GitHub personal access token with at least the `read:packages` scope and access to the Codesphere packages. If the organization enforces SSO, authorize the token for that organization. Use a dedicated machine account where possible so image pulls do not depend on an employee account.
```
walzph: i guess we have no intentions to circle back to the full installer package, right. thus, we should connect access management to our ghcr.io registry to oms itself somehow.

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,439 @@
...
+oms init install-config \
+  --profile production \
+  --config config.yaml \
+  --vault prod.vault.yaml \
+  --with-comments
+```
+
+The profile supplies initial values and the wizard prompts for the final settings. Use `minimal` for the POC topology or `production` for the production baseline. The selected profile configures software defaults; it does not provision or resize machines.
+
+An Ansible inventory can prepopulate the Kubernetes and Ceph host lists. For example:
```
walzph: i'm slightly confused that ansible pops up here as an optional input format, without mentioning it earlier and how it can be used.

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,439 @@
...
+
+Verify every record from outside the private network before continuing.
+
+## 12. Obtain the installer package
+
+Browse the available Codesphere packages and download the selected lite installer build on the jumpbox:
+
+```bash
+oms list packages
+oms download package --version <version> --file installer-lite.tar.gz
```
walzph: does it make sense to default to the lite installer in future again?

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,439 @@
...
+
+```bash
+install -d -m 0700 /etc/codesphere/secrets
+chmod 0600 /etc/codesphere/secrets/age_key.txt
+chmod 0600 /etc/codesphere/secrets/prod.vault.yaml
+test -r /etc/codesphere/secrets/age_key.txt
+test -r /etc/codesphere/secrets/prod.vault.yaml
+```
+
+Run the installation from the jumpbox. The lite package does not contain the platform container images, so skip `load-container-images`; the cluster pulls them directly from GHCR using the credentials in the vault:
```
walzph, line 379: I think that we require `--skip-steps load-container-images` is just a bug. The installer script should know that it can skip this for the lite installation if configured here:
```
registry:
  loadContainerImages: false
```

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,439 @@
...
+
+```bash
+oms install codesphere \
+  --config /etc/codesphere/config.yaml \
+  --priv-key /etc/codesphere/secrets/age_key.txt \
+  --vault /etc/codesphere/secrets/prod.vault.yaml \
+  --package <downloaded-installer-lite-package>.tar.gz \
+  --skip-steps load-container-images
```
walzph, line 387: nit: `skip-steps` could be explained a bit more elaborately. Not every possible value, but why this exists in general.

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,439 @@
...
+- `ceph.nodesSubnet` matches the private host network.
+- `ceph.hosts` lists three hosts for a POC or at least four hosts for production, with exactly one initial master.
+- `ceph.csiKubeletDir` is `/var/lib/k0s/kubelet` when Codesphere manages k0s.
+- Each Ceph OSD definition selects only the intended, empty data and DB/WAL devices.
+- `kubernetes.apiServerHost` is the first control-plane node's private IP, or a stable load-balancer address or DNS name when using multiple control-plane nodes.
+- `kubernetes.controlPlanes` and `kubernetes.workers` assign every k0s node to its intended role. The production baseline uses three dedicated control-plane nodes and a separate worker pool:
+
+  ```yaml
+  kubernetes:
+    managedByCodesphere: true
```
walzph, line 286: Is this the current best practice? I remember SRE telling me they install it externally using k0sctl because of k8s version update issues. This information can be outdated though.
And maybe for external users, it's still better to document the intended way instead of the temporary workaround.

docs/install-codesphere-on-any-infrastructure.md
clemensfreitagdev: Need to double-check, but I think I fixed that bug also 😅

walzph: nope, fails for me on csa demo cluster:
```
Loading container image postgres of component database. 
open /etc/codesphere/deps/database/images/postgres.tar: no such file or directory
```

walzph: still open 😅 
https://github.com/codesphere-cloud/codesphere-monorepo/pull/16318

clemensfreitagdev: 🫢 I try to rebase it tmrw

clemensfreitagdev: Ah that's even the larger change with redefining the props. 

I have another one that adds no need flag but also open 🤯

https://github.com/codesphere-cloud/codesphere-monorepo/pull/17934

walzph: No, don't think this PR can be merged as it is. It's introducing a breaking config change.
I've moved the corresponding CU task to the PE board for triage. 
Tell me if you disagree that this is an OMS task.

clemensfreitagdev: So you think https://github.com/codesphere-cloud/codesphere-monorepo/pull/17934 is also breaking?

walzph: nope that one looks better :) We let Manuel decide, I'd say.

docs/install-codesphere-on-any-infrastructure.md
NautiluX: We're working on it. But it's not there yet, so doesn't make sense to document a desired future state

docs/install-codesphere-on-any-infrastructure.md
NautiluX: it's the case already

docs/install-codesphere-on-any-infrastructure.md
NautiluX: we have a k0sctl based installation in OMS, but not sure if and how it can be used currently. I'll collect input from the team.

docs/install-codesphere-on-any-infrastructure.md
NautiluX: it's there in case someone (like us) uses ansible to configure their hosts

docs/install-codesphere-on-any-infrastructure.md
NautiluX: Not sure what you mean by this, can you propose an edit?

docs/install-codesphere-on-any-infrastructure.md
NautiluX: turned that into a side note.

docs/install-codesphere-on-any-infrastructure.md
NautiluX: added a table to explain them

docs/install-codesphere-on-any-infrastructure.md
NautiluX: I guess we can completely remove that step from the installer since we don't plan to support the full package anymore. For now I tend to leave it as is in case we need to circle back for whatever reason.

docs/install-codesphere-on-any-infrastructure.md
NautiluX: turned it into a side note.

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+      - ipAddress: 10.10.0.16
+  ```
+
+  Always list at least three addresses under `controlPlanes` for a highly available control plane. Dedicated worker addresses go under `workers`; increase that list for the expected workload CPU, memory, and failure tolerance.
+
+  The interactive `oms init install-config` wizard asks separately for the comma-separated control-plane and worker IPs.
+
+  k0s supports combined control-plane/worker nodes. However, the current OMS k0sctl configuration generator treats an IP present in both lists as control-plane-only and ignores the duplicate worker entry. Until combined-role generation is supported by OMS, use dedicated control-plane and worker entries in this workflow.
+- The platform and public gateways use `LoadBalancer`, or `ExternalIP` where that is the chosen integration.
+- `codesphere.domain` is `cs.<base-domain>`.
```
clemensfreitagdev, line 304:
```suggestion
- `codesphere.domain` is `<base-domain>`, can be a sub-domain like `codesphere.corp42.net`
```

docs/install-codesphere-on-any-infrastructure.md
walzph: Agree, but shouldn't we fix the --skip-step bug at least? 
https://github.com/codesphere-cloud/codesphere-monorepo/pull/17934 should be non breaking and we can still control the load-container-images step by using the existing config option.

docs/install-codesphere-on-any-infrastructure.md
walzph: The problem is that the `codesphere.workspaceHostingBaseDomain` parameter was never really configured to something else than `<dataCenter.id>.<codesphere.domain>` at all.
And nothing in the guide tells you that the value must differ between multiple DCs (e.g. by prefixing the datacenter id).
IMO we should 
1. make this parameter optional and default to `<dataCenter.id>.<codesphere.domain>` if undefined
and
2. document the requirement adequately

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+
+## 1. Collect the required inputs
+
+Decide on the following before provisioning anything:
+
+- A base domain whose DNS zone you can change.
+- A datacenter ID, name, city, and country code.
+- Stable private IP addresses for every machine.
+- A public IP address for the jumpbox. This is separate from the service addresses and should accept SSH only from trusted administrator networks.
+- Three stable, externally reachable service IP addresses: platform gateway, workspace gateway, and workspace SSH proxy.
```
joka134, line 39: Should we give some examples how we do it or mention the CCM here?

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+- Access to the Codesphere package portal.
+- An initial cluster administrator email, if desired.
+- OAuth credentials for each Git provider or OIDC provider that users will use.
+- An SSH key with administrative access to all hosts.
+
+Git provider credentials must be supplied as complete sets. For example, a GitHub App needs its app name, client ID, and client secret; an OIDC provider needs its issuer URL, client ID, and client secret.
+
+## 2. Provision the machines
+
+Provision x86-64 machines running Ubuntu 22.04 LTS or an equivalent supported operating system. Choose a topology based on the purpose of the installation.
```
joka134: What's blocking us from recommending newer versions? 🫣

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+sysctl fs.inotify.max_user_watches
+sysctl fs.inotify.max_user_instances
+sysctl vm.max_map_count
+```
+
+Keep time synchronized on every host. Confirm that hostnames, private addresses, attached disks, DNS resolution, outbound connectivity, and SSH access survive a reboot before installing Codesphere.
+
+## 7. Prepare the jumpbox
+
+Install OMS on the jumpbox together with `sops` and `age`. Pin and verify approved versions so the environment is repeatable.
```
joka134: Link to oms installation guide? Especially useful if we move this doc out of github

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+  ```
+
+  Always list at least three addresses under `controlPlanes` for a highly available control plane. Dedicated worker addresses go under `workers`; increase that list for the expected workload CPU, memory, and failure tolerance.
+
+  The interactive `oms init install-config` wizard asks separately for the comma-separated control-plane and worker IPs.
+
+  k0s supports combined control-plane/worker nodes. However, the current OMS k0sctl configuration generator treats an IP present in both lists as control-plane-only and ignores the duplicate worker entry. Until combined-role generation is supported by OMS, use dedicated control-plane and worker entries in this workflow.
+- The platform and public gateways use `LoadBalancer`, or `ExternalIP` where that is the chosen integration.
+- `codesphere.domain` is `cs.<base-domain>`.
+- `codesphere.workspaceHostingBaseDomain` and the custom-domain CNAME base are `ws.<base-domain>`.
```
clemensfreitagdev:
```suggestion
- `codesphere.workspaceHostingBaseDomain`: recommended to set to `<dc-id>.<base-domain>', e.g. `1.codesphere.corp42.net`
- `codesphere.customDomains.cNameBaseDomain`: recommended to set to  `workspaces.<base-domain>`.
```

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+All profiles start with the same common infrastructure defaults, including Codesphere-managed Kubernetes, installed PostgreSQL, Ceph, `LoadBalancer` gateways, disabled MetalLB, and placeholder localhost addresses. They also start with the standard workspace and hosting plan defaults. Replace all placeholder topology, domain, registry, and secrets values during the wizard or configuration review.
+
+Use `minimal` for the POC topology described in section 2 and `production` for the production baseline. Use `dev` only when disabling the bundled monitoring components is appropriate.
+
+> **Optional Ansible inventory:** Users who already maintain host topology in an Ansible inventory can import it with `--ansible-inventory inventory.yaml`. OMS reads the optional `k8s-cp`, `k8s-workers`, and `ceph` groups; each host needs a `private_ip`. For example:
+>
+> ```yaml
+> k8s-cp:
+>   hosts:
+>     cp-1: {private_ip: 10.10.0.11}
```
joka134:
```suggestion
>     cp-1: 
         private_ip: 10.10.0.11
```
No brackets in yaml inventory

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+> k8s-workers:
+>   hosts:
+>     worker-1: {private_ip: 10.10.0.21}
```
joka134:
```suggestion
>     worker-1: 
          private_ip: 10.10.0.21
```

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+> ceph:
+>   hosts:
+>     ceph-1: {private_ip: 10.10.0.31}
```
joka134:
```suggestion
>     ceph-1:
         private_ip: 10.10.0.31
```

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+
+Create these records after the three stable addresses are allocated:
+
+| Record | Target |
+| --- | --- |
+| `cs.<base-domain>` | Platform gateway IP |
+| `*.cs.<base-domain>` | Platform gateway IP |
+| `ws.<base-domain>` | Workspace gateway IP |
+| `*.ws.<base-domain>` | Workspace gateway IP |
+| `*.ssh.cs.<base-domain>` | Workspace SSH proxy IP |
```
joka134: We should stick to the names of the kubernetes services to make it simple to identify.
These are `gateway-controller`, `public-gateway-controller` and `ssh-workspace-proxy`

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+5. Installs and configures Ceph.
+6. Installs and configures k0s when Kubernetes is Codesphere-managed.
+7. Installs Argo CD, cluster dependencies, and managed-service backends.
+8. Installs the Codesphere platform.
+
+The installation can also be run as the separate `infra`, `dependencies`, and `platform` phases when operational change control requires distinct checkpoints.
+
+## 14. Complete the infrastructure integration
+
+If the selected infrastructure needs a Kubernetes cloud controller, install its supported provider integration and enable external cloud-provider mode on the k0s controller and workers. Use only manifests and service annotations intended for that infrastructure.
```
joka134, line 438: cloud controller could be related to the DNS step

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+
+Confirm that the three services receive the reserved addresses:
+
+```bash
+/etc/codesphere/deps/kubernetes/files/k0s kubectl \
+  get services --all-namespaces
```
joka134:
```suggestion
  get services -n codesphere
```
Could grep for LoadBalancer as well.

docs/install-codesphere-on-any-infrastructure.md
```diff
@@ -0,0 +1,450 @@
...
+
+Complete these checks before handing over the environment:
+
+1. All k0s nodes are `Ready`.
+2. Ceph reports healthy and all expected OSDs are present on all three POC hosts or at least four production hosts.
+3. The three load-balancer services have the reserved external addresses.
+4. `https://cs.<base-domain>` presents a trusted certificate and loads Codesphere.
+5. A workspace can be created, reached over HTTPS, and reached through the workspace SSH proxy.
+
+Run the Codesphere smoke test when credentials for the environment are available:
```
joka134:
```suggestion
Run the Codesphere smoke test when an API key for the environment is available:
```

docs/install-codesphere-on-any-infrastructure.md
joka134: Simply link to OMS I guess: https://github.com/codesphere-cloud/oms

docs/install-codesphere-on-any-infrastructure.md
NautiluX: json is yaml no?

docs/install-codesphere-on-any-infrastructure.md
NautiluX: updated

docs/install-codesphere-on-any-infrastructure.md
NautiluX: I think this discussion is independent of the documentation

docs/install-codesphere-on-any-infrastructure.md
NautiluX: keeping the TS k0s for now, the oms-based installation is not yet integrated in the automated testing. Probably the recommendation would anyways be to use k0sctl to install one outside of the TS installer and set to managed.

docs/install-codesphere-on-any-infrastructure.md
NautiluX: updated

docs/install-codesphere-on-any-infrastructure.md
NautiluX: added a minimum versions section

docs/install-codesphere-on-any-infrastructure.md
NautiluX: added

docs/install-codesphere-on-any-infrastructure.md
NautiluX: probably for provider-specific subpages at a later point?

docs/install-codesphere-on-any-infrastructure.md
NautiluX: not sure what you mean

joka134: Depending on your setup you don't get external IPs before setting up the cloud controller. So the DNS step explained before won't work.

NautiluX: added cross-references between sections

joka134 (approved):

PR #582, opened 2026-07-17
feat(fga): add optional openfga db backup secrets

NJona:
Those secrets are required for the openfga db backup that stores it into a s3 bucket.


OliverTrautvetter (approved):
🚀

PR #592, opened 2026-07-21
feat(gcp): support local OMS binary on jumpbox

NautiluX:
Sometimes we want to use a specific version of the OMS binary when bootstrapping a test environment, e.g. one that is built locally or a specific version downloaded in a GitHub action.


internal/bootstrap/gcp/gcp.go
```diff
@@ -843,16 +849,22 @@ func (b *GCPBootstrapper) EnsureJumpboxConfigured() error {
...
+		if err := b.Env.Jumpbox.NodeClient.CopyFile(b.Env.Jumpbox, b.Env.RemoteOmsBinaryPath, "/usr/local/bin/oms"); err != nil {
+			return fmt.Errorf("failed to copy local OMS binary to jumpbox: %w", err)
+		}
+		if err := b.Env.Jumpbox.RunSSHCommand("root", "chmod 0755 /usr/local/bin/oms"); err != nil {
+			return fmt.Errorf("failed to make local OMS binary executable on jumpbox: %w", err)
+		}
+	} else if !b.Env.Jumpbox.HasCommand("oms") {
+		if err := b.Env.Jumpbox.InstallOms(); err != nil {
+			return fmt.Errorf("failed to install OMS on jumpbox: %w", err)
+		}
```
joka134: you can always execute this check even after you did the installation just to verify if it works

internal/installer/node/node.go
```diff
@@ -214,6 +212,34 @@ func (n *Node) InstallOms() error {
...
+// EnsureOmsDependencies installs the command-line tools used by remote OMS
+// workflows independently of how the OMS binary itself was provisioned.
+func (n *Node) EnsureOmsDependencies() error {
+	dependencies := []struct {
+		command string
+		install string
+	}{
+		{
+			command: "sops",
+			install: "curl -LO https://github.com/getsops/sops/releases/download/v3.11.0/sops-v3.11.0.linux.amd64; sudo mv sops-v3.11.0.linux.amd64 /usr/local/bin/sops; sudo chmod +x /usr/local/bin/sops",
```
joka134, line 224: Can we get latest here as well?

go.mod
```diff
@@ -3,6 +3,11 @@ module github.com/codesphere-cloud/oms
 go 1.26.5
 
 replace (
+	// GoReleaser pulls github.com/chrismellard/docker-credential-acr-env,
+	// which imports github.com/Azure/azure-sdk-for-go/version. Azure SDK
+	// v68 removed that package, so keep the legacy monorepo on a verified
+	// version that still provides it.
+	github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v67.2.0+incompatible
```
joka134, line 10: 🫠 
Well, https://pkg.go.dev/github.com/azure/azure-sdk-for-go sticks to an even older version

joka134 (changes requested):
See comments. Also we decided to always have some short PR descriptions as OMS is public and other teams would like to understand the reason for PR's as well

internal/installer/node/node.go
NautiluX: you mean update the sops version? Guess we can do that independently

go.mod
NautiluX: this is shitty, but I don't see another way to solve it for now

internal/bootstrap/gcp/gcp.go
NautiluX: not sure what you mean, you're suggesting to check if the command exists every time? I wouldn't do that because it may hide a problem with copying the local binary

joka134: What do you mean? Aren't we expecting that `b.Env.Jumpbox.HasCommand("oms")` is successful after we copy the local binary?

NautiluX: yes, but this check is used for downloading oms if it's not present, not a validation

joka134: Ok, and my suggestion was to add one by executing this afterwards 😄
but fine for now, it will fail when using oms later on

joka134 (approved):

internal/bootstrap/gcp/gcp.go
NautiluX: refactored the logic to extract OMS installation into a dedicated function

joka134 (approved):

PR #593, opened 2026-07-21
feat(gcp): add ACME staging option

NautiluX:
Allow using the let's encrypt ACME endpoint for bootstrapping to avoid rate limits


joka134 (approved):
Looks good. Please add a short PR description before merging

PR #594, opened 2026-07-22
feat(openfga): add openfga backup config to config.yaml

NJona:
Allows to configure openfga backups in the config.yaml.

Thread regarding config schema for that: https://codesphere-cloud.slack.com/archives/C01Q12KG2Q4/p1784641065573559

Follow up PR adds it to the bootstrap command to test it with each Test PC Installation:
- https://github.com/codesphere-cloud/oms/pull/595

### DIFF
--- a/internal/installer/files/config_yaml.go
+++ b/internal/installer/files/config_yaml.go
@@ -317,6 +317,7 @@ type CodesphereConfig struct {
 	OAuth                      *OAuthProvidersConfig  `yaml:"oauth,omitempty"`
 	ManagedServices            []ManagedServiceConfig `yaml:"managedServices,omitempty"`
 	OpenBao                    *OpenBaoConfig         `yaml:"openBao,omitempty"`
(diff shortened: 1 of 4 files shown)


internal/installer/files/config_yaml.go
```diff
@@ -339,6 +340,21 @@ type OpenBaoConfig struct {
 	User   string `yaml:"user,omitempty"`
 }
 
+// OpenfgaBackupsConfig is the friendly representation of the OpenFGA database
+// backup settings. On marshal it is translated into the openfga subchart values
+// under codesphere.override (see buildOpenfgaBackupOverride).
+type OpenfgaBackupsConfig struct {
```
NautiluX, line 346: shouldn't we also generate this part of the config in init install-config?

NautiluX (approved):
As far as I understand, this PR doesn't do anything since OpenfgaBackups will always be nil, is that intended?

internal/installer/files/config_yaml.go
NJona: Added it to the next PR 👍

PR #595, opened 2026-07-22
feat(fga): add openfga backup to bootstrap

NJona:
Stacked PR, only merge after #594.

Adds openfga backup to the bootstrap, so the backup configuration is tested with each PC Installation. Creates a new bucket in gcp in the same project and configures that as backup storage for openfga. On cleanup the bucket is deleted together with the whole project deletion.

### DIFF
--- a/internal/bootstrap/gcp/install_config.go
+++ b/internal/bootstrap/gcp/install_config.go
@@ -458,9 +458,10 @@ func (b *GCPBootstrapper) applySshProxyConfig() {
 // database backups against the S3-compatible Cloud Storage endpoint.
 const openfgaBackupSAName = "openfga-backup"
 
-// EnsureOpenfgaBackupBucket creates the Cloud Storage bucket, dedicated service
-// account and HMAC key used for OpenFGA database backups. The bucket and HMAC key
-// live in the project so they are removed together with the project on cleanup.
+// EnsureOpenfgaBackupBucket creates the Cloud Storage bucket and HMAC key used for
+// OpenFGA database backups. The bucket and HMAC key live in the project so they are
+// removed together with the project on cleanup. The dedicated service account and
+// its storage role are provisioned by EnsureServiceAccounts / EnsureIAMRoles.
 //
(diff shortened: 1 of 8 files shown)


NautiluX (changes requested):
literally requesting changes I guess :D

internal/bootstrap/gcp/install_config.go
```diff
@@ -453,6 +454,72 @@ func (b *GCPBootstrapper) applySshProxyConfig() {
...
+	}
+	b.Env.OpenfgaBackupBucket = bucketName
+
+	saEmail, _, err := b.GCPClient.CreateServiceAccount(b.Env.ProjectID, openfgaBackupSAName, openfgaBackupSAName)
+	if err != nil {
+		return fmt.Errorf("failed to ensure openfga backup service account: %w", err)
+	}
+	if err := b.GCPClient.AssignIAMRole(b.Env.ProjectID, openfgaBackupSAName, b.Env.ProjectID, []string{"roles/storage.objectAdmin"}); err != nil {
+		return fmt.Errorf("failed to assign storage role to openfga backup service account: %w", err)
+	}
```
NautiluX: I think we have a different function creating service accounts. Not saying it's better to have them all in 1 function, but to have it consistently, I think this should move into the same function, or reuse an existing service account if there is no strict requirement to use another one.

NautiluX (commented):
one comment, else looks good :)

internal/bootstrap/gcp/install_config.go
NJona: Moved the openfga-backup SA creation into `EnsureServiceAccounts` and its `storage.objectAdmin` binding into `EnsureIAMRoles` (using the shared retry helper), so all SAs/roles are provisioned in one place. Kept it as a dedicated SA rather than reusing `cloud-controller` to avoid granting storage admin to the compute SA. `EnsureOpenfgaBackupBucket` now just handles the bucket + HMAC key and derives the SA email deterministically.

NautiluX (approved):

PR #602, opened 2026-07-23
feat(gcp): install k0s with k0sctl before Codesphere

NautiluX:
Use k0sctl installation integrated in OMS to bootstrap on GCP
* Installs k0s using k0sctl and configures it so Codesphere can be installed
* Skips the kubernetes installation of the TS installer
* Sets the default k0s version to current lowest supported Kubernetes version

### DIFF
--- a/internal/bootstrap/gcp/gcp.go
+++ b/internal/bootstrap/gcp/gcp.go
@@ -351,6 +351,16 @@ func (b *GCPBootstrapper) Bootstrap() error {
 	}
 
 	if b.Env.InstallVersion != "" || b.Env.InstallLocal != "" {
+		err = b.stlog.Step("Install k0s", b.InstallK0s)
+		if err != nil {
+			return fmt.Errorf("failed to install k0s: %w", err)
+		}
+
+		err = b.stlog.Step("Wait for k0s nodes", b.WaitForK0sNodes)
+		if err != nil {
+			return fmt.Errorf("failed waiting for k0s nodes: %w", err)
+		}
+
 		err = b.stlog.Step("Install Codesphere", b.InstallCodesphere)
 		if err != nil {
 			return fmt.Errorf("failed to install Codesphere: %w", err)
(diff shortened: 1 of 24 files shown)


internal/bootstrap/gcp/gcp.go
```diff
@@ -1058,52 +1073,103 @@ func (b *GCPBootstrapper) EnsureDNSRecords() error {
...
 }
 
-func (b *GCPBootstrapper) ensureCodespherePackageOnJumpbox() (string, error) {
-	packageFilename := "installer.tar.gz"
+// InstallK0s deploys k0s with the native OMS installer and stores its
+// kubeconfig in the encrypted install vault for the remaining installer steps.
+func (b *GCPBootstrapper) InstallK0s() error {
+	// Reuse matching cached binaries and let k0sctl reconcile normally. Without
+	// --force, an unchanged cluster remains untouched on bootstrap retries.
+	installCmd := fmt.Sprintf("oms install k0s --install-config /etc/codesphere/config.yaml --vault %s --vault-priv-key %s/age_key.txt",
```
joka134: When rerunning this, it fails with:

> Using latest k0s version: v1.36.2+k0s.0
Downloading k0s version v1.36.2+k0s.0
Error: failed to download k0s: k0s binary already exists at /root/.cache/oms/k0s. Use --force to overwrite

So we are not using the version we defined before and we should force it (or remove this behaviour in oms when the binary already exists)

internal/installer/k0sctl_config.go
```diff
@@ -123,12 +123,26 @@ func GenerateK0sctlConfig(installConfig *files.RootConfig, k0sVersion string, ss
...
+	// Track added IPs to avoid emitting the same host twice.
 	addedIPs := make(map[string]bool)
 
-	// Add controller-only nodes from control planes
+	// Add control-plane nodes, enabling their worker role when configured.
 	for _, cp := range installConfig.Kubernetes.ControlPlanes {
-		host := createK0sctlHost(cp, "controller", nil, sshKeyPath, k0sBinaryPath)
+		var installFlags []string
+		if workerIPs[cp.IPAddress] {
+			installFlags = []string{"--enable-worker", "--no-taints=true"}
```
joka134, line 137: I tried setting up a single node cluster on gcp with this branch and the cp was not showing up as worker node. Need some further investigation why.

internal/bootstrap/gcp/gcp.go
NautiluX: done

internal/installer/k0sctl_config.go
NautiluX: this is a nice-to-have follow-up, as discussed

internal/bootstrap/gcp/gcp.go
```diff
@@ -351,6 +351,21 @@ func (b *GCPBootstrapper) Bootstrap() error {
 	}
 
 	if b.Env.InstallVersion != "" || b.Env.InstallLocal != "" {
+		err = b.stlog.Step("Install k0s", b.InstallK0s)
```
joka134, line 354: fails on second run on existing cluster:
```
Downloading k0s version v1.31.14+k0s.0
Error: failed to download k0s: k0s binary already exists at /root/.cache/oms/k0s. Use --force to overwrite
```

joka134 (changes requested):
Found this while testing. Do we always want to force it in the bootstrapper?

internal/bootstrap/gcp/gcp.go
NautiluX: as discussed, I think you missed using the local binary to run the install on the jumpbox. I tried running bootstrapping twice and it worked for me.

joka134 (approved):

PR #607, opened 2026-07-23
refac(cmd): Subpackage for k0s commands

NautiluX:
* Move k0s-related commands to k0s subpackage
* Move sops availability util to package testutil


joka134 (approved):
🚀

PR #608, opened 2026-07-23
feat: deploy argocd and pc-apps in local bootstrap

schrodit:
Enables argo to be deployed by default in the local bootstrap command since multiple components now depend on it.

Moves the ArgoCdAndApps installer to a common pckage that it can be also used in the local bootstrap command


internal/installer/argocd_and_apps.go
joka134, line 1: Can this be a new package in the installer folder? We would like to split the growing installer package into smaller ones and having a package `installer/argocd` would fit into that

cli/cmd/bootstrap_local.go
```diff
@@ -95,7 +95,7 @@ func AddBootstrapLocalCmd(parent *cobra.Command) {
 	flags.StringVar(&bootstrapLocalCmd.CodesphereEnv.InstallConfigPath, "install-config", "", "Path to install config file (default: <install-dir>/config.yaml)")
 	flags.StringVar(&bootstrapLocalCmd.CodesphereEnv.SecretsFilePath, "secrets-file", "", "Path to secrets file (default: <install-dir>/prod.vault.yaml)")
 	// ArgoCD integration
-	flags.BoolVar(&bootstrapLocalCmd.CodesphereEnv.UseArgoCD, "argocd", false, "After infra setup: install ArgoCD, update the OCI pull secret, and install pc-apps from the BOM version")
+	flags.BoolVar(&bootstrapLocalCmd.CodesphereEnv.UseArgoCD, "argocd", true, "After infra setup: install ArgoCD, update the OCI pull secret, and install pc-apps from the BOM version")
 	flags.StringVar(&bootstrapLocalCmd.CodesphereEnv.ArgoCDRegistryURL, "registry-url", "oci://ghcr.io/codesphere-cloud/charts", "OCI registry URL used for the ArgoCD helm pull secret (only relevant with --argocd)")
```
joka134, line 99: > only relevant with --argocd

should rephrase it, as it's the new default. This also means the registry-user is not optional anymore, right?

internal/bootstrap/local/local.go
```diff
@@ -659,9 +674,15 @@ func (b *LocalBootstrapper) UpdateInstallConfig() (err error) {
 		return fmt.Errorf("failed to encrypt vault file: %w", err)
 	}
 
-	creator := vault.NewVaultSecretCreator(b.kubeClient)
-	if err := creator.CreateSecretFromVault(b.ctx, b.icg.GetVault(), vault.VaultSecretNamespace, vault.VaultSecretName); err != nil {
-		return fmt.Errorf("failed to create vault secret: %w", err)
+	if !b.Env.UseArgoCD {
+		// The ArgoCD-and-apps installer owns this sync when that integration is enabled.
+		if err := secrets.EnsureServiceAccountTokens(b.icg.GetVault()); err != nil {
```
joka134: can you move this into a reusable function? Else this has to be maintained twice

joka134 (changes requested):

cli/cmd/bootstrap_local.go
schrodit: it was also not optional before?

internal/bootstrap/local/local.go
schrodit: ah yes, missed to remove it. its already part of the installer

cli/cmd/bootstrap_local.go
joka134: Right. The description says 

> Custom Registry username (optional)

So we should adjust that

schrodit: yes, done.
Its also already marked as required below

joka134 (approved):

PR #609, opened 2026-07-23
refac(cmd): apikey subpackage

NautiluX:
* Move api key handling into apikey subpackage
* Move smoketest to codesphere subpackage


joka134 (approved):
lgfcai

PR #622, opened 2026-07-30
fix(cli): keep updated install vault encrypted

NautiluX:
## Summary

- keep prod.vault.yaml SOPS-encrypted after oms update install-config
- centralize encrypted vault output in InstallConfigManager and reuse the existing SOPS encryption helper
- route temporary-file creation, writes, permissions, cleanup, and replacement through the FileIO abstraction
- replace the encrypted target only after successful encryption, preserving the original vault on failure
- share --postgres-server interpretation between init and update: primary hostname in install mode, server address in external mode
- update generated command documentation

## Tests

- go test ./internal/util ./internal/installer ./internal/installer/vault ./cli/cmd -count=1
- success coverage verifies the updated vault remains encrypted and decryptable
- failure coverage uses in-memory file operations and verifies the original vault is unchanged

ClickUp: https://app.clickup.com/t/24560134/869e71neh


cli/cmd/update_install_config.go
```diff
@@ -199,16 +203,27 @@ func (c *UpdateInstallConfigCmd) applyUpdates(config *files.RootConfig, vault *f
 }
 
 func (c *UpdateInstallConfigCmd) applyPostgresUpdates(config *files.RootConfig, tracker *SecretDependencyTracker) {
```
NautiluX, line 207: shouldn't this be shared code between initial generation and update to avoid drifting functionality?

cli/cmd/update_install_config.go
```diff
@@ -229,12 +244,29 @@ func (c *UpdateInstallConfigCmd) applyPostgresUpdates(config *files.RootConfig,
...
+	vaultData, err := icg.GetVault().Marshal()
+	if err != nil {
+		return fmt.Errorf("failed to marshal vault.yaml: %w", err)
+	}
+	if c.Opts.WithComments {
+		vaultData = installer.AddVaultComments(vaultData)
+	}
+
+	return installervault.EncryptDataWithSOPSAtomically(vaultData, c.Opts.VaultFile, recipient)
+}
```
NautiluX: I think this should be a function of the InstallConfigManager

cli/cmd/update_install_config.go
```diff
@@ -229,12 +244,29 @@ func (c *UpdateInstallConfigCmd) applyPostgresUpdates(config *files.RootConfig,
...
 
+func (c *UpdateInstallConfigCmd) writeEncryptedVault(icg installer.InstallConfigManager) error {
+	recipient, _, err := installervault.ResolveAgeKey("", filepath.Dir(c.Opts.VaultFile))
+	if err != nil {
+		return fmt.Errorf("failed to resolve age key: %w", err)
+	}
+
+	vaultData, err := icg.GetVault().Marshal()
+	if err != nil {
+		return fmt.Errorf("failed to marshal vault.yaml: %w", err)
```
NautiluX: this should mention the actual filename, not vault.yaml

internal/installer/vault/vault_encryption.go
```diff
@@ -165,6 +166,45 @@ func EncryptFileWithSOPS(src, target, recipient string) error {
 	return nil
 }
 
+// EncryptDataWithSOPSAtomically encrypts data into a temporary file next to
+// target and replaces target only after encryption succeeds. Plaintext is
+// passed to SOPS through stdin and is never written to disk.
+func EncryptDataWithSOPSAtomically(data []byte, target, recipient string) error {
```
NautiluX: we already have  EncryptFileWithSOPS in internal/installer/vault/vault_encryption.go and should reuse it

NautiluX: Also, the name Atomically is super non-speaking and doesn't tell what the function does

NautiluX: Removed the additional encryption helper. WriteEncryptedVault now reuses vault.EncryptFileWithSOPS; the manager method name describes the operation.

cli/cmd/update_install_config.go
NautiluX: Moved encrypted vault output to InstallConfigManager.WriteEncryptedVault.

cli/cmd/update_install_config.go
NautiluX: The shared marshaling error now includes the actual target filename via filepath.Base(vaultPath).

cli/cmd/update_install_config.go
NautiluX: Shared the mode-aware Postgres server mapping via determinePostgresServerConfig, which is now used by both init and update.

internal/installer/config_manager.go
```diff
@@ -267,27 +269,96 @@ func (g *InstallConfig) WriteInstallConfig(configPath string, withComments bool)
...
+	if err := vault.EncryptFileWithSOPS(plainPath, encryptedPath, recipient); err != nil {
+		return err
+	}
+
+	mode := os.FileMode(0600)
+	if info, err := os.Stat(vaultPath); err == nil {
+		mode = info.Mode().Perm()
+	} else if !os.IsNotExist(err) {
+		return fmt.Errorf("failed to inspect encrypted vault target: %w", err)
+	}
```
NautiluX: we should be able to formulate this without the need for an else if

internal/installer/config_manager_test.go
```diff
@@ -510,6 +512,29 @@ var _ = Describe("ConfigManager", func() {
...
+			manager := installer.NewInstallConfigManager()
+			err = manager.WriteEncryptedVault(vaultPath, false)
+			Expect(err).To(HaveOccurred())
+
+			content, readErr := os.ReadFile(vaultPath)
+			Expect(readErr).NotTo(HaveOccurred())
+			Expect(content).To(Equal(original))
+		})
+	})
+
```
NautiluX, line 555: we should be able to do this without actual file operations. we already have util/filewriter.go which abstracts these operations. We should use it in the actual function so we're able to test without actually creating files.

internal/installer/config_manager.go
```diff
@@ -267,27 +269,96 @@ func (g *InstallConfig) WriteInstallConfig(configPath string, withComments bool)
...
+	if err := os.Chmod(encryptedPath, mode); err != nil {
+		return fmt.Errorf("failed to set encrypted vault permissions: %w", err)
+	}
+	if err := os.Rename(encryptedPath, vaultPath); err != nil {
+		return fmt.Errorf("failed to replace encrypted vault: %w", err)
+	}
```
NautiluX: use g.FileIO instead of the os package

internal/installer/config_manager.go
```diff
@@ -267,27 +269,96 @@ func (g *InstallConfig) WriteInstallConfig(configPath string, withComments bool)
...
+	}()
+	if _, err := plainFile.Write(vaultYAML); err != nil {
+		_ = plainFile.Close()
+		return fmt.Errorf("failed to write temporary plaintext vault: %w", err)
+	}
+	if err := plainFile.Close(); err != nil {
+		return fmt.Errorf("failed to close temporary plaintext vault: %w", err)
+	}
+
+	encryptedFile, err := os.CreateTemp(filepath.Dir(vaultPath), "."+filepath.Base(vaultPath)+".encrypted-*")
```
NautiluX: probably we need to abstract this in the filewriter.go file so that we can use it here

NautiluX: Added CreateTemp and Rename to util.FileIO/FilesystemWriter and regenerated the FileIO mock.

internal/installer/config_manager.go
NautiluX: WriteEncryptedVault now uses the manager's FileIO for temporary files, writes, cleanup, chmod, and replacement.

internal/installer/config_manager.go
NautiluX: Removed the conditional entirely: encrypted vault permissions are now set directly to 0600.

internal/installer/config_manager_test.go
NautiluX: The test now uses the in-memory MockFileIO plus injected age-key resolution and encryption, so it performs no real file operations.

internal/installer/config_manager.go
```diff
@@ -42,25 +42,47 @@ type InstallConfigManager interface {
...
+	WriteEncryptedVault(vaultPath string, withComments bool) error
 }
 
 type InstallConfig struct {
-	fileIO util.FileIO
-	Config *files.RootConfig
-	Vault  *files.InstallVault
+	fileIO         util.FileIO
+	vaultEncryptor func(src, target, recipient string) error
+	ageKeyResolver func(explicitKeyFile, fallbackDir string) (recipient, keyPath string, err error)
```
NautiluX: use structs and interfaces instead of function types

NautiluX: Replaced both injected function fields with VaultEncryptor and AgeKeyResolver interfaces, concrete SOPS-backed structs, and struct-based test doubles.

cli/cmd/update_install_config.go
```diff
@@ -105,6 +106,7 @@ func AddUpdateInstallConfigCmd(update *cobra.Command, opts *util.GlobalOptions)
 	c.cmd.Flags().StringVar(&c.Opts.PostgresReplicaIP, "postgres-replica-ip", "", "Replica PostgreSQL server IP")
 	c.cmd.Flags().StringVar(&c.Opts.PostgresReplicaName, "postgres-replica-name", "", "Replica PostgreSQL server name")
 	c.cmd.Flags().StringVar(&c.Opts.PostgresServerAddress, "postgres-server-address", "", "PostgreSQL server address (for external mode)")
+	c.cmd.Flags().StringVar(&c.Opts.PostgresServer, "postgres-server", "", "PostgreSQL primary hostname for install mode or server address for external mode")
```
joka134: `--postgres-server-address` and `--postgres-server` sound very similar and from the description it's not very clear how to use it. Can we make this more clear for users or add some examples?

joka134: And I think the flags should be aligned with the one in `init install-config` if possible.

internal/installer/config_manager.go
```diff
@@ -42,25 +42,74 @@ type InstallConfigManager interface {
 	GenerateSecrets() error
 	WriteInstallConfig(configPath string, withComments bool) error
 	WriteVault(vaultPath string, withComments bool) error
+	WriteEncryptedVault(vaultPath string, withComments bool) error
```
joka134: for `LoadVaultFromFile` and `LoadVaultFromUnecryptedFile` we use it exactly the other way around. encrypted is default and we have a special function for unencrypted vaults

internal/installer/config_manager.go
```diff
@@ -42,25 +42,74 @@ type InstallConfigManager interface {
...
+}
+
+type VaultEncryptor interface {
```
joka134: Shouldn't this be part of the vault package?

internal/installer/config_manager.go
```diff
@@ -42,25 +42,74 @@ type InstallConfigManager interface {
...
+// SetAgeKeyResolver overrides age key resolution (useful for testing).
+func (g *InstallConfig) SetAgeKeyResolver(resolver AgeKeyResolver) {
+	g.ageKeyResolver = resolver
+}
+
+func (g *InstallConfig) encryptVault(src, target, recipient string) error {
+	if g.vaultEncryptor != nil {
+		return g.vaultEncryptor.Encrypt(src, target, recipient)
+	}
+	return sopsVaultEncryptor{}.Encrypt(src, target, recipient)
```
joka134: Do we need to cover this case where the encryptor is not set? As they are initialized in the New function, it should be fine to return an error here if that happens for some unexpected reason

joka134 (changes requested):

cli/cmd/update_install_config.go
NautiluX: Updated in 5bc7ac5. `--postgres-server` is now the single visible flag in both `init install-config` and `update install-config`. Its help text explains the install/external behavior, and the generated docs include examples for both modes. The old specific flags remain accepted but are deprecated and hidden for compatibility.

internal/installer/config_manager_test.go
```diff
@@ -25,6 +27,18 @@ type MockFileIO struct {
...
+func (failingVaultEncryptor) Encrypt(_, _, _ string) error {
+	return errors.New("encryption failed")
+}
+
+type staticAgeKeyResolver struct{}
+
+func (staticAgeKeyResolver) Resolve(_, _ string) (string, string, error) {
+	return "recipient", "", nil
+}
```
NautiluX: use a generated mock instead of this

internal/installer/config_manager_test.go
```diff
@@ -510,6 +536,39 @@ var _ = Describe("ConfigManager", func() {
...
+				Vault:  &files.InstallVault{},
+			}
+			manager.SetFileIO(NewMockFileIO())
+			manager.SetAgeKeyResolver(staticAgeKeyResolver{})
+			manager.SetVaultEncryptor(nil)
+
+			err := manager.WriteVault("prod.vault.yaml", false)
+			Expect(err).To(MatchError("vault encryptor is not configured"))
+		})
+	})
```
NautiluX: don't think we need this test

internal/installer/config_manager.go
NautiluX: Changed in 5bc7ac5. `WriteVault` is now the encrypted default, matching `LoadVaultFromFile`. Initial generation/bootstrap paths explicitly use `WriteUnencryptedVault`.

internal/installer/config_manager.go
NautiluX: Yes, moved in 5bc7ac5. The `Encryptor`, `AgeKeyResolver`, and their SOPS/default implementations now live in the `vault` package; the config manager only consumes those interfaces.

internal/installer/config_manager.go
NautiluX: Agreed. The fallback was removed in 5bc7ac5. `NewInstallConfigManager` initializes both dependencies, and an unexpected nil encryptor/resolver now returns an explicit error. I added one focused test for the missing-encryptor case.

cli/cmd/update_install_config.go
```diff
@@ -101,10 +104,13 @@ func AddUpdateInstallConfigCmd(update *cobra.Command, opts *util.GlobalOptions)
 
 	// PostgreSQL update flags
 	c.cmd.Flags().StringVar(&c.Opts.PostgresPrimaryIP, "postgres-primary-ip", "", "Primary PostgreSQL server IP")
-	c.cmd.Flags().StringVar(&c.Opts.PostgresPrimaryHostname, "postgres-primary-hostname", "", "Primary PostgreSQL server hostname")
+	c.cmd.Flags().StringVar(&c.Opts.PostgresPrimaryHostname, "postgres-primary-hostname", "", "Primary PostgreSQL server hostname (deprecated: use --postgres-server)")
```
joka134, line 107: Are we going to remove them later?

joka134 (approved):
🚀

cli/cmd/update_install_config.go
NautiluX: "later" I guess 😬

PR #623, opened 2026-07-31
feat(installer): config and vault primitives for multiple data centers

NJona:
Groundwork for multi-data-center support, with no behaviour change for existing callers.

- `InstallVault.RemoveSecret` and `InstallVault.Clone`, so a vault can be derived from another one instead of generated from scratch.
- `RootConfig.Clone`, a YAML round-trip deep copy.
- The installer's `dataCenters` and `defaultDataCenterId` keys, which declare the topology of the installation. Both are `omitempty`, so a single-data-center config serialises exactly as before.
- `InstallConfigManager.SetInstallConfig` / `SetVault`, so a caller can seed a manager from a source other than a profile or a file.
- `files.PostgresUserSecretName` / `PostgresPasswordSecretName`, so the per-service postgres secret names have a single definition.

## Review notes

Nothing consumes any of this yet — it is all additive.

---

### DIFF
--- a/internal/installer/files/config_yaml.go
+++ b/internal/installer/files/config_yaml.go
@@ -4,6 +4,7 @@
 package files
 
 import (
+	"fmt"
 	"strings"
 
 	"go.yaml.in/yaml/v3"
@@ -39,6 +40,40 @@ func (v *InstallVault) SetSecret(entry SecretEntry) {
 	v.Secrets = append(v.Secrets, entry)
 }
 
+// RemoveSecret deletes the entry with the given name, if present.
+func (v *InstallVault) RemoveSecret(name string) {
+	for i := range v.Secrets {
+		if v.Secrets[i].Name == name {
+			v.Secrets = append(v.Secrets[:i], v.Secrets[i+1:]...)
+			return
+		}
+	}
+}
+
+// Clone returns a deep copy of the vault, so callers can derive a new vault without
+// aliasing the original's secret entries.
+func (v *InstallVault) Clone() *InstallVault {
+	if v == nil {
+		return nil
+	}
+
+	clone := &InstallVault{Secrets: make([]SecretEntry, 0, len(v.Secrets))}
+	for _, entry := range v.Secrets {
+		copied := SecretEntry{Name: entry.Name}
+		if entry.File != nil {
+			file := *entry.File
+			copied.File = &file
+		}
+		if entry.Fields != nil {
+			fields := *entry.Fields
+			copied.Fields = &fields
+		}
+		clone.Secrets = append(clone.Secrets, copied)
+	}
+
+	return clone
+}
+
 func (v *InstallVault) Unmarshal(data []byte) error {
 	return yaml.Unmarshal(data, v)
 }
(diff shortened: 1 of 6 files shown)


internal/tmpl/NOTICE
NJona, line 1: Don't know why those got updated here by the automation. 🤔

joka134: We can't run this in a renovate PR as renovate would ignore the PR from then on after some manual changes. So they are disabled for now. If renovate did an update, the next PR will actually update the docs and licenses 🙃

joka134: But updating to unknown does not seem right 😅

internal/installer/files/config_yaml.go
```diff
@@ -39,6 +40,40 @@ func (v *InstallVault) SetSecret(entry SecretEntry) {
 	v.Secrets = append(v.Secrets, entry)
 }
 
+// RemoveSecret deletes the entry with the given name, if present.
+func (v *InstallVault) RemoveSecret(name string) {
```
joka134: Can we return an error or bool to indicate if something was actually deleted?

joka134: And some unittests for this function would be great

internal/installer/files/config_yaml.go
```diff
@@ -39,6 +40,40 @@ func (v *InstallVault) SetSecret(entry SecretEntry) {
...
+	for i := range v.Secrets {
+		if v.Secrets[i].Name == name {
+			v.Secrets = append(v.Secrets[:i], v.Secrets[i+1:]...)
```
joka134: What about https://pkg.go.dev/golang.org/x/exp/slices#Delete ?

internal/installer/files/config_yaml.go
```diff
@@ -39,6 +40,40 @@ func (v *InstallVault) SetSecret(entry SecretEntry) {
...
+			return
+		}
+	}
+}
+
+// Clone returns a deep copy of the vault, so callers can derive a new vault without
+// aliasing the original's secret entries.
+func (v *InstallVault) Clone() *InstallVault {
```
joka134, line 56: Same here, let's add some tests

internal/installer/files/config_yaml.go
```diff
@@ -39,6 +40,40 @@ func (v *InstallVault) SetSecret(entry SecretEntry) {
...
+	if v == nil {
+		return nil
+	}
```
joka134, line 59: How is v.Clone called if v is nil?

internal/installer/files/config_yaml.go
```diff
@@ -628,6 +670,23 @@ func (c *RootConfig) Marshal() ([]byte, error) {
 	return yaml.Marshal(c)
 }
 
+// Clone returns a deep copy of the config, so a caller can derive a new config without sharing
+// the original's maps, slices and pointers. It round-trips through YAML, so — like any other
+// load of this config — keys the struct does not model are not carried over.
+func (c *RootConfig) Clone() (*RootConfig, error) {
```
joka134, line 694: unit tests? 😄

joka134 (changes requested):

internal/installer/files/config_yaml.go
NJona: Now returns bool for whether an entry was removed.

internal/installer/files/config_yaml.go
NJona: Pointer-receiver methods are callable on nil pointers in Go, so `v.Clone()` on a nil `*InstallVault` returns nil instead of panicking. That lets a caller pass on a vault that was never loaded without guarding every call site. Documented on the method and covered by a test.

joka134 (approved):
🚀 🥇

PR #624, opened 2026-07-31
feat(installer): derive a data center's vault from the primary's

NJona:
A multi-data-center installation shares one PostgreSQL server, so the postgres roles, the token signing keys and everything that encrypts rows in the shared database have to be **identical** in every data center's vault. Everything tied to a single cluster — the ingress CA, the cephadm SSH key, the kubeconfig, the Ceph credentials the installer writes back — must **not** be.

`DeriveDataCenterVault` copies the primary data center's vault and drops exactly the second group, so running `EnsureSecrets` over the result regenerates those and leaves the shared ones untouched. `ClearDataCenterScopedConfig` resets the config fields written alongside a dropped secret, since the `Ensure*` functions are gated on the vault entry and would otherwise leave a stale value paired with a freshly generated key.

## Review notes

- The Ceph cluster credentials are matched **by prefix** (`ceph`, `csi`, `rgw`), because the closed-source installer owns those names and adds to them.
- `postgresPrimaryServerKeyPem` / `postgresReplicaServerKeyPem` are deliberately kept: dropping them would push `regeneratePostgresCerts` into its regeneration branch and mint a server key the running postgres has never seen.
- Pure and GCP-independent, which is where the test coverage sits. No caller yet.
---

### DIFF
--- a/internal/installer/secrets/datacenter.go
+++ b/internal/installer/secrets/datacenter.go
@@ -4,6 +4,7 @@
 package secrets
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/codesphere-cloud/oms/internal/installer/files"
@@ -53,6 +54,9 @@ func IsDataCenterScopedSecret(name string) bool {
 // DeriveDataCenterVault returns a copy of the primary data center's vault with all
 // data-center-scoped secrets removed. Running EnsureSecrets over the result regenerates exactly
 // those, while the shared secrets stay byte-identical across data centers.
+//
+// Derive the config with DeriveDataCenterConfig before running EnsureSecrets, so the config
+// fields paired with the dropped secrets are regenerated along with them.
 func DeriveDataCenterVault(primary *files.InstallVault) *files.InstallVault {
 	derived := primary.Clone()
 	if derived == nil {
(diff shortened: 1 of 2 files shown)


internal/installer/secrets/datacenter.go
```diff
@@ -0,0 +1,87 @@
...
+	derived.Secrets = kept
+
+	return derived
+}
+
+// ClearDataCenterScopedConfig resets the config fields that are written by the Ensure* functions
+// alongside a data-center-scoped vault secret. Those pairs must always be mutated together: the
+// Ensure* functions are gated on the vault entry, so a config field left in place would keep a
+// stale value paired with a freshly generated key.
+func ClearDataCenterScopedConfig(config *files.RootConfig) {
```
joka134: Does it mean we call it in that order?
1. DeriveDataCenterVault
2. Ensure
3. ClearDataCenterScopedConfig

Alternative is to modify ensure so that it's only generating the needed certs?

joka134 (approved):
🚀

internal/installer/secrets/datacenter.go
NJona: Not quite,  both derives run in the same step, before any generation: seedSecondaryDataCenter fills in only what the data center didn't load from its own files, then GenerateSecrets runs in updateInstallConfig. 

Your instinct was right that the naming implied an ordering contract, so I turned it into DeriveDataCenterConfig, the counterpart of DeriveDataCenterVault, it clones the primary's config and resets the data-center-scoped fields in one call, and the clearing is no longer separately callable. On the alternative: `Ensure*` already generates only what's needed, since each one returns early when its vault entry exists, dropping the entry is the regenerate signal, so an explicit list would duplicate that knowledge in a second place.

joka134: cool, thanks for the explanation!

PR #625, opened 2026-07-31
refac(gcp): model the data centers of a bootstrap

NJona:
Until now a bootstrapped project was implicitly a single data center: its nodes, gateway IPs, config paths and domains all lived directly on `CodesphereEnvironment`. Multi-DC needs more than one of each, so this introduces the `DataCenter` type holding everything that must differ per data center, while project-level state (project, VPC, jumpbox, shared postgres node, registry) stays on the environment.

`BuildDataCenters` derives the layout from the flags: one entry today, and with `--multi-dc` a second one that shares the first's PostgreSQL server. The primary data center keeps an empty resource-name suffix, so every name, path and domain a single-DC bootstrap produces is unchanged.

## Review notes

Nothing consumes the layout yet — the callers are migrated in the following PRs. Two mechanisms keep that migration safe:

- `ensureDataCenters` derives the layout on first use and adopts state a caller passed through the legacy top-level environment fields, so every entry point works whether or not `Bootstrap` ran first — including infra files written before multi-DC support.
- `mirrorPrimaryDataCenter` projects the primary data center back onto those fields before the infra file is written, so `cleanup` and `restart-vms` keep reading what they always have. The projection is one-way and never read back.
---

### DIFF
--- a/internal/bootstrap/gcp/datacenter.go
+++ b/internal/bootstrap/gcp/datacenter.go
@@ -1,0 +1,177 @@
+// Copyright (c) Codesphere Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package gcp
+
+import (
+	"fmt"
+
+	"github.com/codesphere-cloud/oms/internal/bootstrap/datacenter"
+	"github.com/codesphere-cloud/oms/internal/installer"
+)
+
+// BuildDataCenters derives the data center layout from the bootstrap environment: a single
+// entry in single-DC mode, and two entries in multi-DC mode where the second one shares the
+// first one's PostgreSQL server.
+func BuildDataCenters(env *CodesphereEnvironment, newICG func() installer.InstallConfigManager) []*datacenter.DataCenter {
+	if !env.MultiDC {
+		// A single data center keeps honouring --datacenter-id. In multi-DC mode the IDs are
+		// derived instead, because they drive the per-data-center domains; validateMultiDC
+		// rejects the combination.
+		id := env.DatacenterID
+		if id == 0 {
+			id = datacenter.PrimaryID
+		}
+
+		return []*datacenter.DataCenter{newDataCenter(env, id, "", newICG)}
+	}
+
+	return []*datacenter.DataCenter{
+		newDataCenter(env, datacenter.PrimaryID, "", newICG),
+		newDataCenter(env, datacenter.PrimaryID+1, "-dc2", newICG),
+	}
+}
+
+// ensureDataCenters makes sure the environment has a usable data center layout. It derives the
+// layout on first use and gives every data center an install config manager, so any entry point
+// works whether or not Bootstrap ran first.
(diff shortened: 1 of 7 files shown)


internal/bootstrap/gcp/datacenter.go
```diff
@@ -0,0 +1,268 @@
...
+
+// primaryDatacenterID is the ID of the first data center. It stays 1 in both modes so
+// single-DC bootstraps keep their existing dataCenter.id.
+const primaryDatacenterID = 1
+
+// DataCenter holds the state of one Codesphere data center inside a bootstrapped GCP project.
+// Project-level state (the GCP project, VPC, jumpbox, shared postgres node and container
+// registry) lives on CodesphereEnvironment; everything that must differ between data centers
+// lives here.
+type DataCenter struct {
```
joka134: I think we can store this somewhere outside of the gcp package, as it does not contain any gcp reference

internal/bootstrap/gcp/datacenter.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+		if i == 0 && b.icg != nil {
+			dc.SetConfigManager(b.icg)
+			continue
+		}
+		dc.SetConfigManager(newICG())
+	}
+}
+
+// adoptLegacyEnvFields moves state that a caller supplied through the legacy top-level
```
joka134: Are you planning to remove the legacy structs afterwards?

joka134 (changes requested):
looks good, but two questions

internal/bootstrap/gcp/datacenter.go
```diff
@@ -0,0 +1,268 @@
...
+		// rejects the combination.
+		id := env.DatacenterID
+		if id == 0 {
+			id = primaryDatacenterID
+		}
+		return []*DataCenter{newDataCenter(env, id, "", newICG)}
+	}
+
+	return []*DataCenter{
+		newDataCenter(env, primaryDatacenterID, "", newICG),
```
joka134: Wouldn't it be more consistent to use the suffix for dc1 as well?

NJona: I wanted to keep it backwards compatible when only installing one dc, that's why I kept the names for them

internal/bootstrap/gcp/datacenter.go
NJona: Yes, and it now happens inside the stack. The mirror was a migration scaffold: PRs 4–7 still read Env.ControlPlaneNodes etc., so it has to stay through those. The new PR 11 (multi-dc-11-drop-legacy-env-mirror) deletes mirrorPrimaryDataCenter and its 8 call sites once nothing reads those fields. The fields themselves stay read-only — ensureDataCenters adopts them so an infra file from a pre-multi-DC OMS still works for cleanup/restart-vms — and got omitempty so new infra files don't write dead keys.

joka134 (approved):

PR #626, opened 2026-07-31
feat(gcp): derive the VM definitions per data center

NJona:
The VM list was a package-level global describing one data center. It becomes a function of the environment: the project-shared jumpbox and postgres VMs plus each data center's three Ceph and three k0s nodes, whose names carry that data center's suffix. A single-DC bootstrap therefore still gets exactly `ceph-1..3` and `k0s-1..3`.

`EnsureComputeInstances` routes each instance into its own data center by the definition's `DataCenterID` instead of inferring placement from the VM's tag, and sorts each data center's nodes independently — the install config assigns roles by index.

`restart-vms` resolves the valid VM names from the infra file's layout, so a node of a second data center can be restarted by name (`--name k0s-1-dc2`). It now also reads the infra file when `--project-id` and `--zone` are given, best-effort, since only the file knows the layout.

## Review notes

`VMDefsForEnv` falls back to a single unsuffixed data center when the environment carries none, which is what an infra file written by an older OMS looks like.

---

### DIFF
--- a/internal/bootstrap/gcp/gce.go
+++ b/internal/bootstrap/gcp/gce.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/compute/apiv1/computepb"
+	"github.com/codesphere-cloud/oms/internal/bootstrap/datacenter"
 	"github.com/codesphere-cloud/oms/internal/github"
 	"github.com/codesphere-cloud/oms/internal/installer/node"
 	"github.com/codesphere-cloud/oms/internal/util"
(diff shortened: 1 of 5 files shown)


internal/bootstrap/gcp/gce.go
```diff
@@ -24,18 +24,69 @@ type VMDef struct {
...
+func sharedVMDefs() []VMDef {
+	return []VMDef{
+		{Name: "jumpbox", MachineType: "e2-medium", Tags: []string{"jumpbox", "ssh"}, AdditionalDisks: []int64{}, ExternalIP: true},
+		{Name: "postgres", MachineType: "e2-standard-2", Tags: []string{"postgres"}, AdditionalDisks: []int64{}, ExternalIP: true},
+	}
+}
+
+// dataCenterVMDefs returns the Ceph and k0s VMs of one data center. The suffix is empty for the
+// primary data center, so single-DC bootstraps keep the names ceph-1..3 and k0s-1..3.
+func dataCenterVMDefs(dcID int, suffix string) []VMDef {
```
joka134, line 52: It would be fine to always add the datacenter suffix. That would simplify the logic

joka134 (approved):
Looks good. Just one idea to always use a consistent suffix style.

internal/bootstrap/gcp/gce.go
NJona: I wanted to make it backwards compatible

PR #627, opened 2026-07-31
feat(gcp): configure hosts, gateways and installs per data center

NJona:
Migrates the infrastructure steps from the single implicit data center to the layout: root login and host configuration run over every data center's nodes, each data center reserves its own gateway, public gateway and SSH proxy IP under a suffixed name, and each gets its own k0s configuration script patching its own gateway services.

`InstallCodesphere` and `RunK0sConfigScript` loop over the data centers in ascending order and log a step per data center. The order matters once there is more than one: the primary's install creates the database, roles and schema the others reuse, and a k0s script can only patch gateway services an install has already created.

The install command moves into the exported `InstallCommand`, since the CLI prints it for the operator when the bootstrap does not install itself, and it now names the data center's own config, vault and age key.

## Review notes

`EnsureHostsConfigured` also creates `/etc/codesphere/secrets` up front on every node: the installer uploads a data center's age key to that fixed path but only creates its own configured `secrets.baseDir`, so for a data center whose `baseDir` differs the upload target would not exist. Harmless, since nodes belong to exactly one data center.

Behaviour for a single data center is unchanged — same VM names, same IP names, same script, same install command.
---

### DIFF
--- a/internal/bootstrap/gcp/gcp.go
+++ b/internal/bootstrap/gcp/gcp.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"path/filepath"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


internal/bootstrap/gcp/gcp.go
```diff
@@ -110,6 +127,30 @@ func (b *GCPBootstrapper) primaryDC() *datacenter.DataCenter {
...
 }
 
+// allNodes returns every node of the project: the jumpbox, the shared postgres node and all
+// data centers' Ceph and k0s nodes.
+func (b *GCPBootstrapper) allNodes() []*node.Node {
+	nodes := []*node.Node{b.Env.Jumpbox, b.Env.PostgreSQLNode}
+	for _, dc := range b.Env.DataCenters {
+		nodes = append(nodes, dc.ControlPlaneNodes...)
+		nodes = append(nodes, dc.CephNodes...)
+	}
```
joka134: You can call `clusterNodes` here

joka134 (approved):

PR #628, opened 2026-07-31
fix(gcp): give every data center its own DNS records

NJona:
Workspaces resolve per data center, and so does the platform: the frontend asks `<dc-id>.<codesphere.domain>` for the configuration of the data center a workspace lives in. OMS only created `cs.<base-domain>` and its wildcard, both pointing at the first data center's gateway, so with more than one data center the second one's endpoint resolved to the first's gateway, which has no route for that host. Every browser request for it was reset — and since the frontend fetches that config before rendering anything, the whole UI failed.

`EnsureDNSRecords` now creates, per data center, its workspace hosting names and SSH proxy name pointing at its own public gateway and SSH proxy, plus `<dc-id>.cs.<base-domain>` and its wildcard pointing at its own **platform** gateway.

## Review notes

- The per-data-center platform names are only created when there is more than one data center: a single one *is* the primary, which `cs.<base-domain>` already resolves to. So single-DC bootstraps still create exactly the same five records.
- The created records are recorded in the infra file so cleanup deletes exactly those. `DeleteDNSRecordSets` therefore takes the record list instead of a base domain; cleanup falls back to deriving the names for older infra files and for a cleanup driven only by `--project-id`.
- Verified against a live two-data-center instance: adding these two records was what made the second data center reachable.
---

### DIFF
--- a/internal/bootstrap/gcp/gcp.go
+++ b/internal/bootstrap/gcp/gcp.go
@@ -64,21 +64,47 @@ func CheckOMSManagedLabel(labels map[string]string) bool {
 	return exists && value == "true"
 }
 
-// GetDNSRecordNames returns the DNS record names that OMS creates for a given base domain.
-func GetDNSRecordNames(baseDomain string) []struct {
-	Name  string
-	Rtype string
-} {
-	return []struct {
-		Name  string
-		Rtype string
-	}{
+// DNSRecordName identifies a DNS record set that OMS manages.
+type DNSRecordName struct {
+	Name  string `json:"name"`
+	Rtype string `json:"rtype"`
+}
+
+// GetDNSRecordNames returns the DNS record names a single-data-center bootstrap creates for a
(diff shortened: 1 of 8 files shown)


push-multi-dc-stack.sh
joka134, line 1: Was this intentionally checked in?

internal/bootstrap/gcp/gcp.go
joka134, line 1: Is it time to split dns related things into a new file in the gcp package?

push-multi-dc-stack.sh
NJona: No, removed it. 🤦

internal/bootstrap/gcp/gcp.go
NJona: Done, the DNS record handling now lives in dns.go with its standalone test in dns_test.go

joka134 (approved):
🚀 🚀 🚀

PR #642, opened 2026-08-05
Refactor vault handling

NautiluX:
Refactor vault handling to abstract the encryption handling. This will unify handling of vaults and allow to add new vault types in the future.
The TS installer requires an encrypted SOPS vault, so `oms install codesphere` only supports the `sops` type.
`oms init install-config` respects the vault type so allows to output an encrypted vault directly.


internal/installer/vault/plain_vault.go
```diff
@@ -0,0 +1,72 @@
...
+	"strings"
+
+	"github.com/codesphere-cloud/oms/internal/installer/files"
+)
+
+// PlainFileVault stores vault YAML as an unencrypted file.
+type PlainFileVault struct{ options FileOptions }
+
+// NewPlainFileVault returns a vault handler for interacting with a plain (unencrypted) file
+func NewPlainFileVault(opts FileOptions) (*PlainFileVault, error) {
```
joka134, line 20:
```suggestion
func NewPlainFileVault(opts FileOptions) (*Vault, error) {
```
You can return the interface type here

internal/installer/vault/sops_vault.go
```diff
@@ -0,0 +1,183 @@
...
+type SOPSVault struct{ options SOPSOptions }
+
+// SOPSOptions contains configuration specific to the SOPS file backend.
+type SOPSOptions struct {
+	File   FileOptions
+	AgeKey string
+}
+
+// NewSOPSVault creates a new vault handler for interacting with SOPS-encrypted vault files
+func NewSOPSVault(opts SOPSOptions) (*SOPSVault, error) {
```
joka134, line 29:
```suggestion
func NewSOPSVault(opts SOPSOptions) (*Vault, error) {
```

internal/installer/vault/plain_vault.go
joka134, line 20: we could move the implementations of the interface into sub packages to import them like vault.sops.New

internal/installer/vault/vault.go
```diff
@@ -0,0 +1,151 @@
...
+		})
+	case TypePlain:
+		return NewPlainFileVault(FileOptions{Path: opts.Path, WithComments: opts.WithComments, FileIO: opts.FileIO})
+	default:
+		return nil, fmt.Errorf("unsupported vault type %q", vaultType)
+	}
+}
+
+// NewFromString parses vaultType and creates the matching implementation.
+func NewFromString(vaultType string, opts Options) (Vault, error) {
```
joka134, line 97: Why not fore the usage of the custom type?

joka134 (changes requested):
Some minor questions. looks nice in general

internal/installer/vault/sops_vault.go
NautiluX: I think in general it's preferable to return the struct type instead of the interface, makes navigation easier. The caller can use the concrete type and you can jump to that instead of the interface directly.

joka134: That is true, but with this there is no actual verification if this is really implementing the interface. Or am I wrong?

NautiluX: compilation will fail, yes

joka134: because its used in `vault.go` by returning the interface. Not ideal, but I'm fine to keep it for now for convenience during development 👍

NautiluX: huh, let's discuss it on a call briefly, I think it is ideal actually :D

internal/installer/vault/vault.go
NautiluX: I don't understand this comment, can you elaborate?

joka134: Instead of parsing a string we could expect the custom vault type here, so that the caller has to do something like NewFromType(vault.TypeSOPS)

internal/installer/vault/plain_vault.go
NautiluX: done

internal/installer/vault/vault.go
NautiluX: done

internal/installer/vault/plain_vault.go
NautiluX: no! :D

joka134: 😿

joka134 (approved):
nice, lgtm 🚀

PR #652, opened 2026-08-07
fix(bootstrap): Query org members only once

NautiluX:
* Previously we queried them once per VM
* SSH keys are now generated only once and passed to the VM creation

### DIFF
--- a/internal/bootstrap/gcp/gce.go
+++ b/internal/bootstrap/gcp/gce.go
@@ -59,12 +59,16 @@ func (b *GCPBootstrapper) EnsureComputeInstances() error {
 	errCh := make(chan error, len(vmDefs))
 	resultCh := make(chan vmResult, len(vmDefs))
 	logCh := make(chan string, len(vmDefs))
+	sshKeys, err := b.getSSHKeys()
+	if err != nil {
+		return fmt.Errorf("failed to determine SSH keys: %w", err)
+	}
 
 	for _, vm := range vmDefs {
 		wg.Add(1)
 		go func(vm VMDef) {
 			defer wg.Done()
-			result, err := b.ensureVM(vm, b.Env.RootDiskSize, logCh)
+			result, err := b.ensureVM(vm, b.Env.RootDiskSize, sshKeys, logCh)
 			if err != nil {
 				errCh <- err
 				return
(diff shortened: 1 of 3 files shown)


internal/bootstrap/gcp/gce.go
```diff
@@ -59,12 +59,16 @@ func (b *GCPBootstrapper) EnsureComputeInstances() error {
 	errCh := make(chan error, len(vmDefs))
```
OliverTrautvetter, line 59: tests are broken

internal/bootstrap/gcp/gce.go
```diff
@@ -59,12 +59,16 @@ func (b *GCPBootstrapper) EnsureComputeInstances() error {
...
 	resultCh := make(chan vmResult, len(vmDefs))
 	logCh := make(chan string, len(vmDefs))
+	sshKeys, err := b.generateSSHKeys()
```
OliverTrautvetter: Always done now, so also when all VMs already exist and are running. An idempotent re run of bootstrap should fail now

internal/bootstrap/gcp/gce.go
```diff
@@ -122,9 +126,28 @@ func (b *GCPBootstrapper) EnsureComputeInstances() error {
 	return nil
 }
 
+func (b *GCPBootstrapper) generateSSHKeys() (string, error) {
```
OliverTrautvetter: nit, the name is a bit confusing. It does not generate a ssh key, but reads from team / public key

OliverTrautvetter: no unit tests for the new function

OliverTrautvetter (changes requested):
looks good in general, some small comments from my side (tests are broken right now, not sure how big that problem is)

internal/bootstrap/gcp/gce.go
NautiluX: it's a private function, convered as previously by (now failing) tests

internal/bootstrap/gcp/gce.go
```diff
@@ -122,9 +126,28 @@ func (b *GCPBootstrapper) EnsureComputeInstances() error {
...
+	sshKeys := ""
+	if b.Env.GitHubPAT != "" && b.Env.GitHubTeamOrg != "" && b.Env.GitHubTeamSlug != "" {
+		var err error
+		sshKeys, err = github.GetSSHKeysFromGitHubTeam(b.GitHubClient, b.Env.GitHubTeamOrg, b.Env.GitHubTeamSlug)
+		if err != nil {
+			return "", fmt.Errorf("failed to get SSH keys from GitHub team: %w", err)
+		}
+	}
+
+	pubKey, err := b.ReadSSHKey(b.Env.SSHPublicKeyPath)
```
mociarain, line 139: Unrelated but asking to better understand: 

I think ReadSSHKey could (maybe even should) be private. Not a big deal but wondering what you think

mociarain (approved):
LGTM. Left one unrelated question

internal/bootstrap/gcp/gce.go
NautiluX: fixed

internal/bootstrap/gcp/gce.go
NautiluX: why would a re-run fail? It's true in this case they will be queried again.

internal/bootstrap/gcp/gce.go
NautiluX: it's public because it has unit tests. Something to rethink, I agree.

internal/bootstrap/gcp/gce.go
NautiluX: updated

internal/bootstrap/gcp/gce.go
OliverTrautvetter: Right, I meant that a rerun would fail if generateSSHKeys fails to collect the key

NautiluX: that an issue worth turning it into lazy querying of SSH keys? I'm not sure tbh...

OliverTrautvetter: yeah, I am slo not sure how common that issue would be

NautiluX: can we resolve then? I think a rerun can have the same requirements as the first run - these keys must be accessible.

OliverTrautvetter: Yes makes sense

PR #655, opened 2026-08-07
feat: silence usage information on errors

NautiluX:
* Do not print usage information on failures
* Still print usage information on invalid or missing arguments

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -80,7 +80,26 @@ func GetRootCmd() *cobra.Command {
 	// Cluster administration commands
 	AddAddClusterAdminCmd(rootCmd, opts)
 
+	silenceUsageOnRunErrors(rootCmd)
+
 	return rootCmd
+}
+
+// silenceUsageOnRunErrors prevents Cobra from printing command usage for
+// operational errors returned by RunE. Argument, flag, and required-flag
+// validation happens before RunE, so those errors still include usage.
+func silenceUsageOnRunErrors(command *cobra.Command) {
+	if command.RunE != nil {
+		runE := command.RunE
+		command.RunE = func(cmd *cobra.Command, args []string) error {
+			cmd.Root().SilenceUsage = true
+			return runE(cmd, args)
+		}
+	}
+
+	for _, child := range command.Commands() {
+		silenceUsageOnRunErrors(child)
+	}
 }
 
 // Execute executes the root command. This is called by main.main(). It only needs to happen once to the rootCmd.
--- a/cli/cmd/root_usage_test.go
+++ b/cli/cmd/root_usage_test.go
@@ -1,0 +1,66 @@
+// Copyright (c) Codesphere Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"bytes"
+	"errors"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
+)
+
+var _ = Describe("command usage on errors", func() {
+	execute := func(root *cobra.Command, args ...string) (string, error) {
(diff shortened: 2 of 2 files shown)


cli/cmd/root.go
```diff
@@ -80,9 +80,28 @@ func GetRootCmd() *cobra.Command {
...
+			cmd.Root().SilenceUsage = true
+			return runE(cmd, args)
+		}
+	}
+
+	for _, child := range command.Commands() {
+		silenceUsageOnRunErrors(child)
+	}
+}
+
```
OliverTrautvetter, line 104: Several commands validate user input inside RunE. For example in list_packages.go we do:
```
func (c *ListBuildsCmd) RunE(_ *cobra.Command, args []string) error {
	if c.Opts.Sort != portal.SortSemver && c.Opts.Sort != portal.SortDate {
		return fmt.Errorf("invalid sort parameter: %s (must be '%s' or '%s')", c.Opts.Sort, portal.SortSemver, portal.SortDate)
	}
```

So this helpful error would be no longer helpful

cli/cmd/root.go
```diff
@@ -80,9 +80,28 @@ func GetRootCmd() *cobra.Command {
...
+// validation happens before RunE, so those errors still include usage.
+func silenceUsageOnRunErrors(command *cobra.Command) {
+	if command.RunE != nil {
+		runE := command.RunE
+		command.RunE = func(cmd *cobra.Command, args []string) error {
+			cmd.Root().SilenceUsage = true
+			return runE(cmd, args)
+		}
+	}
+
```
OliverTrautvetter, line 99: We also have PreRunE for `download_package.go` that is not handled. We don't do some user stuff in there, but in the future someone could add a PreRunE with user informations that would not be silenced. Is that fine for now or should we already handle all RunE that exist?

```
// PersistentPreRun: children of this command will inherit and execute.
    PersistentPreRun func(cmd *Command, args []string)
    // PersistentPreRunE: PersistentPreRun but returns an error.
    PersistentPreRunE func(cmd *Command, args []string) error
    // PreRun: children of this command will not inherit.
    PreRun func(cmd *Command, args []string)
    // PreRunE: PreRun but returns an error.
    PreRunE func(cmd *Command, args []string) error
    // Run: Typically the actual work function. Most commands will only implement this.
    Run func(cmd *Command, args []string)
    // RunE: Run but returns an error.
    RunE func(cmd *Command, args []string) error
    // PostRun: run after the Run command.
    PostRun func(cmd *Command, args []string)
    // PostRunE: PostRun but returns an error.
    PostRunE func(cmd *Command, args []string) error
    // PersistentPostRun: children of this command will inherit and execute after PostRun.
    PersistentPostRun func(cmd *Command, args []string)
    // PersistentPostRunE: PersistentPostRun but returns an error.
    PersistentPostRunE func(cmd *Command, args []string) error
```

cli/cmd/root.go
NautiluX: what do you mean with no longer useful? That error would still be returned, but not the full usage information is printed (`--help` output)

cli/cmd/root.go
NautiluX: do PreRunE result in the usage being printed?

cli/cmd/root.go
OliverTrautvetter: Yeah I checked the errors again. They are all self-explanatory and the user can as you said just get further information with --help

cli/cmd/root.go
OliverTrautvetter: It should still print it. PreRunE is done before RunE and only in RunE we set `SilenceUsage = true`. But we only have 1 PreRunE in the code, so I don't think this is blocking

OliverTrautvetter (approved):
🚀

PR #656, opened 2026-08-07
feat(openfga): add psk auth to config

NJona:
For multi dc we need to expose openfga from one dc to be reused by all others (one central required component). This can now be done by enabling gateway and setting the host.

Everything in this PR was already validated manually in a multi dc setup.

Related PRs that were used for the testing:
- Adding PSK Secret to Codesphere Services: https://github.com/codesphere-cloud/codesphere-monorepo/pull/20870
- Charts: https://github.com/codesphere-cloud/charts/pull/1135


cli/cmd/update_install_config.go
NJona, line 1: adds missing secrets on `update install-config` command. (Until now it only updated postgres stuff, not missing secrets that can be auto generated)

cli/cmd/update_install_config.go
```diff
@@ -398,6 +408,41 @@ func (c *UpdateInstallConfigCmd) applyCodesphereUpdates(config *files.RootConfig
...
+// in the vault is restored afterwards — an operator's own values are never modified here.
+func addMissingSecrets(config *files.RootConfig, vault *files.InstallVault) ([]string, error) {
+	existing := make(map[string]files.SecretEntry, len(vault.Secrets))
+	for _, secret := range vault.Secrets {
+		existing[secret.Name] = secret
+	}
+
+	if err := secrets.EnsureSecrets(vault, config); err != nil {
+		return nil, fmt.Errorf("ensure secrets: %w", err)
+	}
```
NautiluX, line 537: could we do this the other way round so that we can first fill up the missing secrets on a copy, optionally ask the user to confirm the list of secrets is what they want to add, and then copy them over?

cli/cmd/update_install_config.go
```diff
@@ -182,6 +183,15 @@ func (c *UpdateInstallConfigCmd) UpdateInstallConfig(icg installer.InstallConfig
...
+	added, err := addMissingSecrets(config, vault)
+	if err != nil {
+		return fmt.Errorf("failed to add missing secrets: %w", err)
+	}
+
+	if len(added) > 0 {
+		log.Printf("\nAdded %d secret(s) missing from the vault: %s\n", len(added), strings.Join(added, ", "))
+	}
+
 	if err := icg.WriteInstallConfig(c.Opts.ConfigFile, c.Opts.WithComments); err != nil {
```
NautiluX, line 214: bonus points if the install config ends up printing a diff before writing as well, unless it's completely new ;)

NJona: would do that in separate PR

PR #657, opened 2026-08-07
feat: Add persistant verbose flag

NautiluX:
* hide verbose tar output if not --verbose


internal/installer/codesphere.go
```diff
@@ -179,7 +179,7 @@ func (ci *CodesphereInstaller) warnIfVaultDirDiffersFromSecretsDir(config files.
 }
 
 func (ci *CodesphereInstaller) ExtractAndValidatePackage(pm PackageManager) error {
```
OliverTrautvetter, line 184: false is hardcoded 3 times, can't we add a bool `Verbose` to `CodesphereInstaller`? Otherwise the new --verbose flag ist just ignored here and always false

internal/bootstrap/local/installer.go
```diff
@@ -156,7 +156,7 @@ func (b *LocalBootstrapper) PrepareInstallerBundle() (string, error) {
 	}
 
 	log.Printf("Extracting installer bundle %s → %s", bundlePath, destDir)
-	if err := util.ExtractTarGz(b.fw, bundlePath, destDir); err != nil {
+	if err := util.ExtractTarGz(b.fw, bundlePath, destDir, false); err != nil {
```
OliverTrautvetter: same here, hardcoded false instead of adding a Verbose bool to `LocalBootstrapper`. Flag is ignored

internal/util/tar.go
OliverTrautvetter: prints that are not handled with `csio.Verbosef`

internal/installer/package.go
```diff
@@ -183,7 +183,7 @@ func (p *Package) GetBaseimagePath(baseimage string, force bool) (string, error)
 	}
 
 	baseImageTarPath := path.Join(baseimagePath, baseimage)
-	err := p.ExtractDependency(baseImageTarPath, force)
+	err := p.ExtractDependency(baseImageTarPath, force, false)
```
OliverTrautvetter: hardcoded false instead of using the --verbose flag

internal/util/tar.go
OliverTrautvetter, line 1: no unit tests for the verbose gating

cli/cmd/root.go
```diff
@@ -53,6 +53,9 @@ func GetRootCmd() *cobra.Command {
 			}
 		},
 	}
+
+	rootCmd.PersistentFlags().BoolVarP(&opts.Verbose, "verbose", "", false, "Enable verbose output")
```
OliverTrautvetter, line 57: now there is a global verbose flag, but some commands have a `--quiet` flag. I think we can remove those flags now that we have a global flag.

But maybe something for a follow up pr?

internal/installer/codesphere.go
NautiluX: Added Verbose to CodesphereInstaller and now propagate opts.Verbose through every install phase into Extract and ExtractDependency.

internal/bootstrap/local/installer.go
NautiluX: Added Verbose to LocalBootstrapper, passed the global option in from bootstrap-local, and replaced both hardcoded extraction values.

internal/util/tar.go
NautiluX: Updated openTar, openTarGz, and unsupported-header output to use csio.Verbosef, with the verbose value threaded through the helpers.

internal/installer/package.go
NautiluX: Added verbose parameters to ExtractOciImageIndex and GetBaseimagePath and propagate the global option from the extend/update callers into dependency extraction.

internal/util/tar.go
NautiluX: Added tests covering both sides of tar verbose gating, plus a test verifying CodesphereInstaller propagates Verbose=true to package extraction.

cli/cmd/root.go
NautiluX: Done in this PR: removed the general --quiet flags from package download, k0s download, and smoketests. The downstream quiet arguments now use !Verbose. I kept --ssh-quiet because it specifically controls SSH command output.

cli/cmd/codesphere/install_codesphere.go
```diff
@@ -65,14 +65,17 @@ func (c *InstallCodesphereCmd) RunE(cmd *cobra.Command, _ []string) error {
 	infraInstaller := &installer.CodesphereInstaller{
```
OliverTrautvetter, line 69: 1 more quite flag

cli/cmd/bootstrap_gcp.go
`flags.BoolVar(&bootstrapGcpCmd.SSHQuiet, "ssh-quiet", false, "Suppress SSH command output (default: false)")
`

OliverTrautvetter (approved):
Looks very good now 🚀, approved already. Just 1 comment, I found 1 more quite flag

PR #678, opened 2026-08-12
feat(oms): confirm secret changes in update install-config

NJona:
Two changes to `oms update install-config`:

It now generates the secrets an existing vault is missing. A vault written by an older oms predates whatever the current one requires, and nothing on the upgrade path fills the gap: `oms install` never touches secrets and `oms init install-config` writes a fresh vault rather than extending one. Strictly additive — every entry the vault already holds is kept, including the ones EnsureSecrets would otherwise overwrite.

And it asks before it changes anything in the vault, listing what it would regenerate or generate first. `-y`/`--yes` approves up front; a run without a terminal answers no, so an unattended upgrade never rewrites a secret by itself. Declining a regeneration aborts the update — the certificates cover values that would otherwise already be in the config — while declining a missing secret just skips it.


cli/cmd/update_install_config.go
```diff
@@ -398,6 +427,126 @@ func (c *UpdateInstallConfigCmd) applyCodesphereUpdates(config *files.RootConfig
...
+
+	answer, err := bufio.NewReader(os.Stdin).ReadString('\n')
+	if err != nil && !errors.Is(err, io.EOF) {
+		return false, fmt.Errorf("read stdin: %w", err)
+	}
+
+	answer = strings.TrimSpace(strings.ToLower(answer))
+
+	return answer == "y" || answer == "yes", nil
+}
```
NautiluX: I _think_ we already have tooling for this somewhere around https://github.com/codesphere-cloud/oms/blob/4822985136931183ebd226cb3957e05faf909eff/internal/installer/prompt.go probably we can reuse that or update it to get this functionality as well? Maybe should move it to a different package then, since it's independent of the installation

internal/prompt/prompt.go
```diff
@@ -120,6 +123,13 @@ func (p *Prompter) Bool(prompt string, defaultValue bool) bool {
 	return input == "y" || input == "yes"
 }
 
+// Confirm asks a yes/no question that has to be answered with an explicit yes. Anything
+// else — an empty line, a closed stdin, a non-interactive prompter — counts as no, so a
+// caller can use it to guard a change that must not happen by accident.
+func (p *Prompter) Confirm(prompt string) bool {
+	return p.Bool(prompt, false)
+}
```
NautiluX: do we need this indirection?

cli/cmd/update_install_config.go
```diff
@@ -21,6 +23,10 @@ type UpdateInstallConfigCmd struct {
 	cmd        *cobra.Command
 	Opts       *UpdateInstallConfigOpts
 	FileWriter intutil.FileIO
+
+	// Confirm asks the operator whether to go ahead with a change to the vault. Replaced in
+	// tests; --yes short-circuits it.
+	Confirm func(question string) bool
```
NautiluX: I think it would be more readable (and how I think most of other code does it) to use the Prompter type here instead of a function type. For tests, if you need to replace the behavior, you would create an interface for the prompter and generate a mock.

cli/cmd/update_install_config.go
```diff
@@ -398,6 +421,98 @@ func (c *UpdateInstallConfigCmd) applyCodesphereUpdates(config *files.RootConfig
...
+	for _, item := range items {
+		log.Printf("  - %s\n", item)
+	}
+
+	return c.Confirm(question)
+}
+
+// maybeAddMissingSecrets asks about the secrets the vault is missing and adds them if the
+// operator agrees. Returns the names of the ones that were added.
+func (c *UpdateInstallConfigCmd) maybeAddMissingSecrets(config *files.RootConfig, vault *files.InstallVault) ([]string, error) {
```
NautiluX: function title isn't really speaking about what the function does, it sounds like it was non-deterministic, but why? Is it because it's `interactivelyAddMissingSecrets`?

cli/cmd/update_install_config.go
```diff
@@ -496,6 +614,26 @@ func (t *SecretDependencyTracker) ACMEConfigChanged() bool {
...
+		changes = append(changes, "PostgreSQL primary server certificate")
+	}
+
+	if t.postgresReplicaCertNeedsRegen {
+		changes = append(changes, "PostgreSQL replica server certificate")
+	}
+
+	if t.acmeConfigChanged {
+		changes = append(changes, "ACME configuration")
+	}
```
NautiluX, line 622: I like this human readable output 👍

internal/prompt/prompt.go
NJona: Removed, the caller passes Bool(question, false).

cli/cmd/update_install_config.go
NJona: Prompter is now an interface next to the implementation (concrete type renamed StdinPrompter), with a mockery mock in the tests.

cli/cmd/update_install_config.go
NJona: Renamed to confirmAndAddMissingSecrets, the "maybe" was because it asks first.

cli/cmd/update_install_config.go
```diff
@@ -398,6 +421,99 @@ func (c *UpdateInstallConfigCmd) applyCodesphereUpdates(config *files.RootConfig
...
+
+// addMissingSecrets fills in the secrets an existing vault does not have yet and returns
+// their names. A vault written by an older oms predates whatever the current one requires,
+// and nothing else on the upgrade path generates the difference:
+// `oms install` never touches secrets, and `oms init install-config` writes a fresh vault
+// rather than extending one.
+//
+// Strictly additive. EnsureSecrets is idempotent except for EnsureDefaultSecrets, which
+// always overwrites digitalOceanApiToken with a dummy value, so every entry that was already
+// in the vault is restored afterwards — an operator's own values are never modified here.
```
NautiluX: this is a pretty long comment for a function that's pretty self explanatory. Why does it comment on a flaw in EnsureSecrets (tbh what it describes I think that shouldn't be the case)?

internal/prompt/prompt.go
```diff
@@ -11,19 +14,36 @@ import (
 	"strings"
 )
 
-type Prompter struct {
+// Prompter asks the operator a question and returns their answer, falling back to the
+// default whenever there is none — an empty line, a closed stdin, or a prompter that was
```
NautiluX: so many en-dashs in comments :D

NautiluX (approved):
only two minor things about comments, lgtm otherwise

PR #679, opened 2026-08-12
feat(openfga): add the openFga block to the install config

NJona:
One OpenFGA instance serves a whole installation, so in a multi-data-center setup exactly one data center deploys and publishes it and the others only point at it. That is an installation-level decision rather than a chart detail, so operators express it in config.yaml:

```
  codesphere:
    openFga:
      deploy: true
      apiUrl: https://openfga.1.cs.example.com
      expose:
        enabled: true
        host: openfga.1.cs.example.com
```

Validation rejects the combinations that cannot work: a data center that does not deploy OpenFGA has nowhere to fall back to and must name the instance it uses, an exposed one needs a host, and a data center cannot expose an OpenFGA it does not deploy.

Nothing reads the block yet; deriving the pc-applications values from it follows separately.

### DIFF
--- a/internal/installer/files/config_yaml.go
+++ b/internal/installer/files/config_yaml.go
@@ -361,6 +361,7 @@ type CodesphereConfig struct {
 	ManagedServices            []ManagedServiceConfig `yaml:"managedServices,omitempty"`
 	OpenBao                    *OpenBaoConfig         `yaml:"openBao,omitempty"`
 	OpenfgaBackups             *OpenfgaBackupsConfig  `yaml:"openfgaBackups,omitempty"`
+	OpenFga                    *OpenFgaConfig         `yaml:"openFga,omitempty"`
 	Migration                  *MigrationConfig       `yaml:"migration,omitempty"`
 	TelemetryExport            *TelemetryExport       `yaml:"telemetryExport,omitempty"`
 	Override                   ChartOverride          `yaml:"override,omitempty"`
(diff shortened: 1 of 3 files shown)


internal/installer/files/config_yaml.go
```diff
@@ -398,6 +399,39 @@ type OpenfgaBackupsConfig struct {
...
+	// in-cluster service of a locally deployed OpenFGA; required when Deploy is false.
+	APIURL string `yaml:"apiUrl,omitempty"`
+	// Expose publishes the deployed OpenFGA through the Codesphere gateway so the other
+	// data centers can reach it.
+	Expose *OpenFgaExposeConfig `yaml:"expose,omitempty"`
+}
+
+// OpenFgaExposeConfig publishes a locally deployed OpenFGA through the Codesphere gateway.
+type OpenFgaExposeConfig struct {
+	Enabled bool `yaml:"enabled"`
```
NautiluX, line 419: this will be set to false by default when not set somewhere in the code, is that intended?

NJona: Yes, exposing must be opt-in (only multi-dc), so the zero value is the safe one: no expose block, or enabled omitted, means not published. That's also why Deploy is a *bool and this isn't: deploy defaults to true, so it needs to tell "unset" from "false", while enabled defaults to false and doesn't.

NautiluX (approved):

PR #680, opened 2026-08-12
feat(openfga): derive the pc-applications values

NJona:
Turns the codesphere.openFga block into values for the openfga application: whether pc-applications deploys OpenFGA here, and the gateway that publishes it, with the certificate following the installation's own cert issuer. The derived values are the base — an explicit pcApps block in config.yaml and --pc-apps-values still win over them.

Authentication is not configured but read off the vault: OpenFGA requires the preshared key exactly when the installation has one. An installation whose vault predates the key stays unauthenticated and keeps working, and the Codesphere services decide the same way from the same entry, so the two cannot disagree. Services that started before the key existed pick it up when their pods roll.

### DIFF
--- a/internal/installer/openfga_pc_apps.go
+++ b/internal/installer/openfga_pc_apps.go
@@ -1,0 +1,109 @@
+// Copyright (c) Codesphere Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package installer
+
+import (
+	"github.com/codesphere-cloud/oms/internal/installer/files"
+)
+
+// openFgaPresharedKeysSecret is the Secret the openfga chart reads the preshared key from.
+// Its content is synced out of the installation vault by the chart's own ExternalSecret, so
+// oms only has to name it.
+const openFgaPresharedKeysSecret = "openfga-preshared-keys"
+
+// OpenFgaPcAppsValues translates the customer-facing codesphere.openFga block of the install
+// config into pc-applications values.
+//
+// OpenFGA is deployed by pc-applications, but whether a data center runs its own instance and
+// whether that instance is published is an installation-level decision, not a chart detail — so
+// operators configure it in config.yaml and this derives the chart values from it. The result is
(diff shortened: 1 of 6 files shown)


internal/installer/openfga_pc_apps.go
```diff
@@ -0,0 +1,109 @@
...
+// --pc-apps-values file still override it.
+//
+// Authentication follows the vault rather than the config: OpenFGA requires the preshared key
+// exactly when the installation has one, which keeps it in step with the Codesphere services
+// (they take the same key from the same vault entry, and treat it as optional). Pods that were
+// started before the key existed keep running without it until the release rolls out, so an
+// installation that adds the key mid-life restarts its Codesphere services.
+//
+// Returns nil when neither the config nor the vault says anything about OpenFGA, leaving the
+// pc-applications chart defaults untouched.
```
NautiluX: this is a loooooong comment. Wondering if either the code can explain some of this or if we can make the code more obvious if it's not self-explanatory

internal/installer/openfga_pc_apps.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if fga := config.Codesphere.OpenFga; fga != nil {
+		openfga["enabled"] = fga.DeploysOpenFga()
+
+		if fga.Expose != nil {
+			chartValues["gateway"] = gatewayValues(config, fga.Expose)
+		}
+	}
+
+	if hasOpenFgaPresharedKey(vault) {
```
NautiluX: should we print a notice that the user should set this secret as future versions may require it?

NJona: good idea, done 👍

internal/installer/openfga_pc_apps.go
NJona: Split it up, `authnValues` is its own function now and carries that paragraph (entry point is down to 4 lines).

NautiluX (approved):

PR #697, opened 2026-08-17
Config Templating: Update doc strings and docs with correct quotes

mociarain:
The real work is all here: https://github.com/codesphere-cloud/oms/pull/457

This is just a little update on the docs.

### DIFF
--- a/cli/cmd/template_config.go
+++ b/cli/cmd/template_config.go
@@ -56,14 +56,14 @@ This command prints the rendered configuration to stdout so templating can be te
 Template syntax in config.yaml:
 
   # Inject a secret value (defaults to the "content"/"password" field)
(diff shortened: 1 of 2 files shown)


schrodit (commented):
why correct quotes?
The current double quotes should also work, or is there somthing not working?

mociarain:
> why correct quotes? The current double quotes should also work, or is there somthing not working?

The double, double quotes isn't valid yaml. 

`
echo 'someKey: "{{ secret "mySecret" }}"' | yq .

// Error: bad file '-': yaml: while parsing a block mapping at <unknown position>: <unknown position>: did not find expected key
`

tadas-cs (approved):

PR #708, opened 2026-08-20
feat(cli): add oms copy package command for transferring images between registries

schrodit:
Introduces a new `oms copy` command group with a `package` subcommand that copies container images and OCI Helm charts referenced in a BOM between registries. Adds supporting package-copy logic in internal/installer, extends the BOM model, and adds a trailing newline to bootstrap step/substep log lines so copy progress output doesn't run together.

### DIFF
--- a/cli/cmd/copy_package.go
+++ b/cli/cmd/copy_package.go
@@ -1,0 +1,199 @@
+// Copyright (c) Codesphere Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+
+	csio "github.com/codesphere-cloud/cs-go/pkg/io"
+	"github.com/codesphere-cloud/oms/cli/cmd/util"
+	"github.com/codesphere-cloud/oms/internal/env"
+	"github.com/codesphere-cloud/oms/internal/installer"
+	"github.com/codesphere-cloud/oms/internal/portal"
+	"github.com/codesphere-cloud/oms/internal/prompt"
+	intutil "github.com/codesphere-cloud/oms/internal/util"
+	"github.com/google/go-containerregistry/pkg/logs"
+	"github.com/spf13/cobra"
+)
+
+const defaultInstallerPackageArtifact = "installer-lite.tar.gz"
+
+// CopyPackageCmd represents the copy package command.
+type CopyPackageCmd struct {
+	cmd        *cobra.Command
+	Opts       CopyPackageOpts
+	Env        env.Env
+	FileWriter intutil.FileIO
+	Prompter   prompt.Prompter
+}
+
+// CopyPackageOpts holds the flags accepted by the copy package command.
+type CopyPackageOpts struct {
+	*util.GlobalOptions
+	Package       string
+	Version       string
+	Hash          string
+	Filename      string
+	Dest          string
+	Yes           bool
+	Force         bool
+	Insecure      bool
+	Verbose       bool
+	ShowArtifacts bool
+}
+
+// RunE resolves the installer package and executes its artifact transfers.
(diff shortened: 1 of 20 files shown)


internal/bootstrap/bootstrap_stepper.go
```diff
@@ -59,7 +59,7 @@ func (b *StepLogger) Substep(name string, fn func() error) error {
 	b.subSteps += 1
 	b.currentStep = name
 
-	fmt.Printf("%s%s   %s...", LINE_RESET, RESET_TEXT, name)
+	fmt.Printf("%s%s   %s...\n", LINE_RESET, RESET_TEXT, name)
```
OliverTrautvetter: why do we need this \n ?

The Step function is not called by copy_package.go and package_copy.go. But it changes shared code used by every bootstrap flow

cli/cmd/copy_package.go
```diff
@@ -0,0 +1,199 @@
...
+	if err := installer.CopyPackageArtifacts(ctx, copier, artifacts); err != nil {
+		return fmt.Errorf("failed to copy package artifacts: %w", err)
+	}
+
+	log.Printf("Successfully copied %d package artifacts to %s", len(artifacts), c.Opts.Dest)
+
+	return nil
+}
+
+func (c *CopyPackageCmd) resolvePackage(portalClient portal.Portal) (string, error) {
```
OliverTrautvetter, line 154: Looks like it duplicates
`func (c *DownloadPackageCmd) DownloadBuild(p portal.Portal, build portal.Build, filename string) error {
` from download_package.go

I think it is worth to extract the shared download and verify logic so both commands stay in sync

internal/installer/package_copy.go
```diff
@@ -0,0 +1,140 @@
...
+
+	var craneOutput bytes.Buffer
+	logs.Warn.SetOutput(&craneOutput)
+	logs.Progress.SetOutput(&craneOutput)
+	defer func() {
+		logs.Warn.SetOutput(io.Discard)
+		logs.Progress.SetOutput(io.Discard)
+	}()
+
+	for i, artifact := range artifacts {
```
OliverTrautvetter, line 123: sequential copy loop. Not sure how long it takes max, but maybe some simple worker pool would already improve this. So to do like 4 crane.Copy at a time instead of just 1

cli/cmd/copy_package_test.go
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) Codesphere Inc.
```
OliverTrautvetter, line 1: no test for the --version download path

OliverTrautvetter (changes requested):
Looks good in general, 1 bigger question and some smaller ones

internal/bootstrap/bootstrap_stepper.go
schrodit: was doing some experiments for proper human readable output.
Will remove it again

internal/installer/package_copy.go
schrodit: I also thought about that, but crane itself already does a paralyzation of pushing the layers in parallel.
So I think it does make it any faster to also parallalize on the upper layer since we are restricted by bandwith of the machine.

cli/cmd/copy_package.go
schrodit: Agreed, extracted the shared download+verify logic into portal.DownloadAndVerifyBuild (internal/portal/download.go). Both DownloadPackageCmd.DownloadBuild and CopyPackageCmd.resolvePackage now call it, so they stay in sync (download.go supports resume for the download command, copy just always creates fresh).

cli/cmd/copy_package_test.go
schrodit: Added coverage for the --version path in copy_package_test.go (resolvePackage: success case downloading+verifying via the portal, and the GetBuild failure case).

OliverTrautvetter (approved):
lgtm 🚀

PR #723, opened 2026-08-21
refac(internal/github): single graphql query to fetch all SSH keys at once

gnarlex:
We are making 1 list request to get all members of a team, and 1 get request for each member's ssh key. For large team's, this results in a lot of requests being made, which may quickly contribute to hitting GitHub API rate limits,

Make a single GraphQL query to get all keys at once.

### DIFF
--- a/internal/github/github_client.go
+++ b/internal/github/github_client.go
@@ -4,39 +4,160 @@
 package github
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
 
-	"github.com/google/go-github/v74/github"
 	"golang.org/x/oauth2"
 )
 
-// GitHubClient abstracts the GitHub API calls used to fetch team SSH keys.
+const githubGraphQLEndpoint = "https://api.github.com/graphql"
+
+// teamMemberSSHKeysQuery fetches every member of a team together with their public SSH keys in a
+// single request. Members are paginated with the $after cursor; publicKeys are assumed to fit in
+// the first page (a user is extremely unlikely to have more than 20 keys).
+const teamMemberSSHKeysQuery = `query($org: String!, $team: String!, $after: String) {
+  organization(login: $org) {
+    team(slug: $team) {
+      members(first: 100, after: $after) {
+        pageInfo { hasNextPage endCursor }
+        nodes {
+          login
+          publicKeys(first: 20) { nodes { key } }
+        }
+      }
+    }
+  }
+}`
+
+// TeamMemberKeys holds a team member's login and their public SSH keys.
+type TeamMemberKeys struct {
+	Login string
+	Keys  []string
+}
+
+// GitHubClient abstracts the GitHub API call used to fetch team SSH keys.
 //
 //mockery:generate: true
 type GitHubClient interface {
-	ListTeamMembersBySlug(ctx context.Context, org, teamSlug string, opts *github.TeamListTeamMembersOptions) ([]*github.User, error)
-	ListUserKeys(ctx context.Context, username string) ([]*github.Key, error)
+	GetTeamMemberSSHKeys(ctx context.Context, org, teamSlug string) ([]TeamMemberKeys, error)
 }
 
 type RealGitHubClient struct {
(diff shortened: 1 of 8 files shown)


internal/github/github_client.go
```diff
@@ -4,39 +4,160 @@
...
+		}
+
+		if !team.Members.PageInfo.HasNextPage {
+			break
+		}
+		cursor := team.Members.PageInfo.EndCursor
+		after = &cursor
+	}
+
+	return members, nil
```
OliverTrautvetter, line 128: Doesn't this silently succeed with 0 keys if the team/org is misconfigured? Before it was a hard 404 error, now it would be (nil, nil)

internal/github/github_client.go
```diff
@@ -4,39 +4,160 @@
...
+// GraphQL API, following member pagination until every member has been retrieved.
+func (c *RealGitHubClient) GetTeamMemberSSHKeys(ctx context.Context, org, teamSlug string) ([]TeamMemberKeys, error) {
+	var members []TeamMemberKeys
+	var after *string
+
+	for {
+		resp, err := c.queryTeamMembers(ctx, org, teamSlug, after)
+		if err != nil {
+			return nil, err
+		}
```
OliverTrautvetter, line 106: drops previously fetched members. We could return the partial results if an error happens.

If partial results are usable?

internal/github/github_client.go
```diff
@@ -4,39 +4,160 @@
...
+// single request. Members are paginated with the $after cursor; publicKeys are assumed to fit in
+// the first page (a user is extremely unlikely to have more than 20 keys).
+const teamMemberSSHKeysQuery = `query($org: String!, $team: String!, $after: String) {
+  organization(login: $org) {
+    team(slug: $team) {
+      members(first: 100, after: $after) {
+        pageInfo { hasNextPage endCursor }
+        nodes {
+          login
+          publicKeys(first: 20) { nodes { key } }
```
OliverTrautvetter: Silently only the first 20. We should at least log that it is truncated and keys not just vanish silently

internal/github/github_client.go
```diff
@@ -4,39 +4,160 @@
...
+		return nil, fmt.Errorf("GraphQL request failed with status %d: %s", httpResp.StatusCode, string(respBody))
+	}
+
+	var result graphQLResponse
+	if err := json.Unmarshal(respBody, &result); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal GraphQL response: %w", err)
+	}
+	if len(result.Errors) > 0 {
+		return nil, fmt.Errorf("GraphQL query returned errors: %s", result.Errors[0].Message)
+	}
```
OliverTrautvetter, line 174: Why only the first Error? We should join them to return all of them

internal/github/github_client.go
OliverTrautvetter, line 1: I think you can do a go mod tidy. My vs code says

`github.com/google/go-github/v74 is not used in this module
`

OliverTrautvetter (changes requested):
Looks good in general, some small stuff

internal/github/github_client.go
gnarlex: I think returning a partial result risks giving a false sense of "it worked". Returning all or nothing is IMHO safer.

internal/github/github_client.go
gnarlex: 404 is not what I would expect from a list members endpoint. If the team/org exists, but has no members, it should return an empty list (nil or empty slice usually have the same behavior, but can return an empty slice if that's preferred) and not return an error. If the caller considers the empty list an error, it's the caller's responsibility to flag this to their user (via error or warning log).

Or am I missing an unhandled error somewhere?

OliverTrautvetter: I agree with that, I just wanted to know if that was an intentional change

internal/github/github_client.go
OliverTrautvetter: fine for me

OliverTrautvetter (approved):
lgtm 🚀

